PR #2, opened 2020-03-20
Added biorxiv embeddings and global analysis

danich1:
This PR contains TSNE and UMAP embeddings of BioRxiv. The document embeddings are generated via an average of all word vectors within a document. Interestingly enough Word2Vec is able to identify documents that have arguably questionable labels (discussed in biorxiv_visualize_embeddings.ipynb). When viewing the embedding visualization notebook try and see if github can load the document. If not I'll save the annotated figures as separate files to aid review.


biorxiv/word_vector_experiment/biorxiv_generate_embeddings.py
```diff
@@ -0,0 +1,279 @@
...
+import lxml.etree as ET
+import pandas as pd
+from sklearn.manifold import TSNE
+from tqdm import tqdm_notebook
+import umap
+
+
+# # Output Documents to File
+
+# This section dumps all of biorxiv text into a single document in order to train the word2vec model. This is for ease of training the model.
```
ajlee21: I think we discussed this a little, but just to clarify.  Your training set consists of all sentences from all the documents together.  So you're not keeping track of sentences from the same document?  So you're loosing some context about where the sentences came from and the meaning of words can differ depending on the overall context of the article..Or does this give a more robust representation of the words because they are trained over many different contexts...?

biorxiv/word_vector_experiment/biorxiv_generate_embeddings.py
```diff
@@ -0,0 +1,279 @@
...
+
+    def __iter__(self): 
+        for line in open(self.filepath, "r"): 
+            yield line.split() 
+
+
+# In[ ]:
+
+
+word_embedding_sizes = [150, 250, 300]
```
ajlee21, line 91: This is the number of features used to represent a given word.  So "apple" is represented by a vector of size 150x1 ?

biorxiv/word_vector_experiment/biorxiv_generate_embeddings.py
```diff
@@ -0,0 +1,279 @@
...
+    biorxiv_vec_df.to_csv(
+        f"output/word2vec_output/biorxiv_all_articles_{model_dim}.tsv.xz", 
+        sep="\t", index=False,
+        compression="xz"
+    )
+
+
+# # UMAP the Documents
+
+# After generating document embeddings, the next step is to visualize all the documents. In order to visualize the embeddings a low dimensional representation is needed. UMAP is an algorithm that can generate this representation, while grouping similar embeddings together.
```
ajlee21, line 208: Is this set of cells redundant to the other notebook?

biorxiv/word_vector_experiment/biorxiv_visualize_embeddings.py
```diff
@@ -0,0 +1,362 @@
...
+    + p9.labs(
+        title="UMAP of BioRxiv (300 dim)",
+        color="Article Category"
+    )
+)
+g.save("output/embedding_output/umap/figures/biorxiv_umap_300.png", dpi=500)
+print(g)
+
+
+# Looking at the plots it seems that documents in the same category mostly cluster together, which is expected. The plots appear to be rotated versions of each other, which suggests that dimensionality doesn't hugely affect umap's results. Interesting cases are the outliers that appear within the plot. Question that arises: what are those outliers?
```
ajlee21, line 155: Really cool! I wonder if there's anyway to do some sort of compression step that might help to distinguish groups more. So group combinations of words and then do this clustering?

biorxiv/word_vector_experiment/biorxiv_visualize_embeddings.py
```diff
@@ -0,0 +1,362 @@
...
+    + p9.annotate("rect", xmin=0, xmax=3, ymin=7.5, ymax=9.5, alpha = 0.4)
+    + p9.labs(
+        title="Outlier Detection of BioRxiv UMAP (300 dim)",
+        color="Article Category"
+    )
+)
+print(g)
+
+
+# In[13]:
```
ajlee21: I wonder if it would be at all helpful to have multiple different datasets:
1. simple -- two categories that you expect to be pretty different, epidemiology and bioengineering
2. complex -- categories that are subtopics like cell biology, cancer biology, biochemistry..

I wonder if you can come up with a hierarchy of categories...not sure if this is useful..just a thought

biorxiv/word_vector_experiment/biorxiv_visualize_embeddings.py
```diff
@@ -0,0 +1,362 @@
...
+    biorxiv_tsne_models_lastest[300]
+    .query("tsne1 > -30 & tsne1 < -20")
+    .query("tsne2 > 40 & tsne2 < 55")
+    .query("category != 'neuroscience'")
+    .category
+    .value_counts()
+)
+
+
+# Looking into the bioinformatics blob approximately 370 articles are off category.
```
ajlee21, line 373: So these are articles that are currently classified as "not bioinformatics" and your classification is say they should be "bioinformatics"?

ajlee21 (approved):
Looks super cool!  So excited to see where this goes.  Just some questions to clarify

biorxiv/word_vector_experiment/biorxiv_generate_embeddings.py
danich1: > Your training set consists of all sentences from all the documents together. So you're not keeping track of sentences from the same document? 

Correct. I am training the word2vec model using all sentences in **every** biorxiv document. 

>So you're loosing some context about where the sentences came from and the meaning of words can differ depending on the overall context of the article..Or does this give a more robust representation of the words because they are trained over many different contexts...?

The latter is what is happening. Since I am using every biorxiv document, I can avoid biases such as spurious associations or missing associations. (e.g. two words are associated in a neruoscience context but would have been missed if I stuck with a bioinformatics context.)

biorxiv/word_vector_experiment/biorxiv_generate_embeddings.py
danich1: Bingo.  "apple" is mapped to a 1x150 vector of network weights.

biorxiv/word_vector_experiment/biorxiv_generate_embeddings.py
danich1: >Is this set of cells redundant to the other notebook?

No. Actually, this set of cells saves me time by generating all necessary embeddings for the visualize biorxiv experiment. I grouped all embedding generating code into one notebook and the other notebooks are designed to visualize the embeddings. This design made the most sense in my head.

biorxiv/word_vector_experiment/biorxiv_visualize_embeddings.py
danich1: No classification in this situation just document embeddings. I'm operating under the hypothesis: `documents under the same category should cluster together rather than documents in different categories.`
In layman's terms: blue should cluster with blue instead of blue having random red points. Since that isn't the case, I needed to look more into why, hence the output tables in the notebook.

biorxiv/word_vector_experiment/biorxiv_visualize_embeddings.py
danich1: Hmm this is a good point. Given the scope of this PR I won't generate the plot here. Will keep it in mind though.

biorxiv/word_vector_experiment/biorxiv_visualize_embeddings.py
danich1: >I wonder if there's anyway to do some sort of compression step that might help to distinguish groups more. So group combinations of words and then do this clustering?

Hmmm.. I thought that's what tSNE and UMAP are suppose to do for datapoints (documents in my context). Am I missing something here?

cgreene: PCA might be more interpretable... no guarantees of course. Did you try it?

danich1: Haven't tried it yet, but will do now.

danich1:
Added PCA plots. Seems that there are overlaps between the categories as I'd expect. Good to confirm that the number of dimensions doesn't matter.

PR #3, opened 2020-03-25
Missing Methods Section Experiment

danich1:
This PR contains result of plotting preprints without methods sections against preprints with methods section. The underlying assumption is that preprints without methods sections would have different embeddings than preprints that do have a method section. Based on this notebook the assumption is wrong and will have to think about next steps on identifying those preprints.


ben-heil (approved):
Good analysis! Have you considered training an L1-regularized logistic regression model on like 100 UMAP dimensions? It makes sense to me that the main source of variation in a paper embedding would be subfield specific vocabulary, while things like presence/absence of a methods section might end up in one of the later PCs or UMAP dimensions.

biorxiv/missing_sections_experiment/biorxiv_missing_method_section_experiment.py
```diff
@@ -0,0 +1,624 @@
...
+
+
+import pandas as pd
+import plotnine as p9
+
+
+# In[2]:
+
+
+biorxiv_sections_df = (
```
ajlee21, line 18: Just to clarify, so you're looking for methods sections based on "method" header presence?

biorxiv/missing_sections_experiment/biorxiv_missing_method_section_experiment.py
```diff
@@ -0,0 +1,624 @@
...
+)
+biorxiv_sections_df.head()
+
+
+# # TSNE Embeddings for BioRxiv
+
+# In[3]:
+
+
+biorxiv_embeddings_df = pd.read_csv(
```
ajlee21, line 30: So unlike your previous analysis, here you're embedding a full document as opposed to embedding words? Just curious, how does the document embedding work? Is this also based on the vocab of all the documents?

ajlee21 (approved):
Very nice short analysis

I agree with @ben-heil , I would also expect more variability to come from the differences in the text. Although methods sections tend to have their own distinct structure, more numbers and technical terminology..I wonder why this wasn't detected.  Maybe the effect got washed out by the rest of the document..?

biorxiv/missing_sections_experiment/biorxiv_missing_method_section_experiment.py
danich1: Bingo. If I can't find "method" header or select variations of it, then the document is considered missing a method's section.

biorxiv/missing_sections_experiment/biorxiv_missing_method_section_experiment.py
danich1: >So unlike your previous analysis, here you're embedding a full document as opposed to embedding words? 

Previous analysis had full document embeddings as well. The document embeddings were generated via the process described below.

>Just curious, how does the document embedding work? Is this also based on the vocab of all the documents?

Yup you got it. Pipeline for generating document embeddings:
1. Generate word embeddings
1a. Take all documents and dump text into a single file
1b. Run word2vec on the entire file 
1c. Obtain model that has words mapped to vectors: "word"-> [#, #, #, #, ...]

2. Convert word embeddings to document embeddings
2a. For each document gather every word and obtain their associated vectors
2b. Take an average of all word vectors found in the document
2c. This average is now considered a document embedding
2d. repeat until all documents have been processed.

danich1:
>Although methods sections tend to have their own distinct structure, more numbers and technical terminology..I wonder why this wasn't detected. Maybe the effect got washed out by the rest of the document..?

Great question. I'm going to look further into it, but mainly what I found is that preprint headers can vary a lot. Plus, I do have some errors in my section detection code. Will be iterating and smoothing that part out, which may help fix some false positives. Going to merge this PR though.

PR #4, opened 2020-04-02
PCA associations

danich1:
This PR contains pretty interesting word cloud figures and a sanity check that there isn't a bug with PCA (Plots look the same in the visualize embeddings notebook PR #2). The central idea here is that PCA is able to capture the strong variance signal contained in the document embedding vectors. Plus, PCA is easy to interpret which makes future analysis quite easy. 

Following the PCA run, I calculated the cosine similarity between word vectors and each principal component's eigenvector. This metric is defined as the cosine of the angle between two vectors. This means that a positive number close to 1 signifies that two vectors are close to being on top of each other and a negative number close to -1 signifies that two vectors are far apart. [visual diagram](https://datascience-enthusiast.com/figures/cosine_sim.png) 

Lastly, I generated word clouds containing the top 20 words associated with each eigenvector. Orange represents the positive end (close to 1) and the blue representing the negative end (close to -1).

Main files to review:
biorxiv_pca_check.py
biorxiv_word_pca_component.py
pca(1-9)_cossim_word_cloud.png  - feel free to look at more if you'd like.


ben-heil (approved):
It's interesting that the correlated and anticorrelated portions of each principle component both seem to correspond to a field, but the fields don't necessarily seem like opposites. For example, PCA5 has evolutionary bio on one end and clinical whole exome sequencing on the other.

It is very cool that different fields are falling out along PC dimensions.

Interesting plots:

PC 9 has some strange things going on in the blue words. They look like they're subject headings or something.

PC 20 is the mosquito-borne virus vector (pun not intended but greatly appreciated). 

PC 31 has some plant bio

biorxiv/pca_association_experiment/biorxiv_pca_check.py
```diff
@@ -0,0 +1,151 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# # Determine Word to PCA Associations
+
+# This notebook is designed to check if there is a bug for the PCA plots. Turns out there isn't a bug the signal captured in the first two components is quite strong.
```
ajlee21: Sorry if i missed this, but I'm a bit unclear what the experiment is that you're conducting.  So what was the reason you suspected a bug? How will you tell that there is not a bug in the signal? I guess more generally, I'm wondering what your eval metric and controls are:)

biorxiv/pca_association_experiment/biorxiv_pca_check.py
```diff
@@ -0,0 +1,151 @@
...
+    )
+
+    pca_dict[embedding_dim] = (
+        pd.DataFrame(embedding, columns=[f"pca{dim}" for dim in range(1, n_components+1, 1)])
+        .assign(document=biorxiv_articles_dict[embedding_dim].document.values.tolist())
+        .merge(journal_map_df[["category", "document", "doi"]], on="document")
+    )
+
+
+# # PCA Embedding Check
```
ajlee21, line 88: What are you "checking" here? I assume this is a manual visual inspection, but for what?

biorxiv/pca_association_experiment/biorxiv_pca_check.py
```diff
@@ -0,0 +1,151 @@
...
+        color="Article Category"
+    )
+)
+print(g)
+
+
+# In[15]:
+
+
+g = (
```
ajlee21, line 132: Maybe helpful to have a take-away from these plots.  Looks like PC2 is somewhat distinguishing between none and everything else. Not sure about PC1

biorxiv/pca_association_experiment/biorxiv_word_pca_component.py
```diff
@@ -0,0 +1,280 @@
...
+    for word in tqdm_notebook(model.wv.vocab.keys())
+}
+
+
+# In[5]:
+
+
+word_vector_df = pd.DataFrame.from_dict(word_vector_map, orient='index')
+
+word_vector_df.to_csv(
```
ajlee21, line 58: Just checking my understanding, so this is you taking your words and embedding them into a 300-dim latent space using the trained model embedding (i.e. trained encoder)?

biorxiv/pca_association_experiment/biorxiv_word_pca_component.py
```diff
@@ -0,0 +1,280 @@
...
+    "output/word_pca_similarity/word_vectors_300.tsv.xz",
+    sep="\t", index_col=0
+)
+
+
+# In[7]:
+
+
+# 1 - cosine distance = cosine similarity
+word_pca_similarity = (
```
ajlee21, line 119: I think the comment wording confused me here. 
So you're comparing how similar your embedding word is (say "water" ) with the PC1 vector? Where the PC1 is meant to capture the variance in your documents? 
So doing this, if "water" is very similar to PC1, that would mean that the word "water" is highly contributing to the variance in the documents?

biorxiv/pca_association_experiment/biorxiv_word_pca_component.py
```diff
@@ -0,0 +1,280 @@
...
+            for record in word_class_map[word_class]
+        })
+        .recolor(color_func=polarity_color_map)
+        .to_file(f"output/word_pca_similarity/figures/pca_{pc}_cossim_word_cloud.png")
+    )
+
+
+# # Document Centroid Cosine Similarity
+
+# Finally this section calculates document centroid to principal component associations. This means the higher score the higher the association a document is with a given principal component.
```
ajlee21: Why are you looking at how similar a cluster of documents are with a certain PC?

ajlee21 (approved):
Really nice experiment,

Just some comments to clarify my understanding that might call for added text if you think its necessary

Cool to look at the word clouds...I would've expected PC1 to have more specific language like "neurons" or "genes"...Curious what you think about the clouds

MarvinT:
This looks cool! I'll need to look over this more in-depth in the future.
I just wanted to let you know about https://radimrehurek.com/gensim/models/keyedvectors.html#gensim.models.keyedvectors.WordEmbeddingsKeyedVectors.similar_by_vector if you weren't aware of it. I didn't have time to look through the whole PR to figure out if it would be useful though, sorry.

biorxiv/pca_association_experiment/biorxiv_pca_check.py
danich1: >So what was the reason you suspected a bug?

This notebook was constructed after a meeting I had with Casey. Explanation was in description above, but I have reproduced it below. :) 
>This PR contains pretty interesting word cloud figures and a sanity check that there isn't a bug with PCA (Plots look the same in the visualize embeddings notebook PR #2).

Basically the PCA plots look the same, which lead to the assumption that I was running PCA on the same word2vec model output (300 dim compared to 150, 250 and 300dim). Turns out the captured variance remains the same regardless of embedding dimensions. 

> How will you tell that there is not a bug in the signal?

The only way to tell it isn't a bug is to look at the output. (visual confirmation) I double checked the embeddings were the same dimensions as intended and then ran PCA on each one. If the PCA plots looked different after running on each dimension embedding, then that would signify I have a bug in my previous analysis (not good).

biorxiv/pca_association_experiment/biorxiv_pca_check.py
danich1: Just to make sure the PCA embedding values don't look the same across models. Just a sanity check that I'm not feed the exact same data into the PCA model.

biorxiv/pca_association_experiment/biorxiv_pca_check.py
danich1: Main takeaway from this notebook is that there isn't a bug with PCA. I'll add that text into the notebook to clarify any confusion. I'll leave principal component descriptions for the next notebook, which dives deeper into each principal component.

biorxiv/pca_association_experiment/biorxiv_word_pca_component.py
danich1: Yup. These words were determined by the word2vec model that I trained in #2. So basically the model has words -> 300dim vectors and I'm just grabbing each word and vector.

biorxiv/pca_association_experiment/biorxiv_word_pca_component.py
danich1: Yup you got. The main idea here is to see what words are contributing to the variance of the documents. A high positive value means that word has a large role in pushing documents up on the positive axes of the principal component. A high negative value means that word has a large role in pushing documents towards the negative axis of principal component.

biorxiv/pca_association_experiment/biorxiv_word_pca_component.py
danich1: This section was a result of me jumping off of the idea that I should examine the association of document centroids to principal components. I agree that this section needs to be fleshed out more, but I was trying to work ahead and have code that does the above.

danich1:
>Cool to look at the word clouds...I would've expected PC1 to have more specific language like "neurons" or "genes"...Curious what you think about the clouds

Good observation. Turns out that the greatest variance (PC1) between documents is words that distinguish quantitative biology and molecular biology. Words such as "neurons" and "genes" appear in PC2 (second greatest variance), which depicts genomics vs neuroscience. Neuroscience appearing in early PCs makes sense, given that majority of bioRxiv consist of neuroscience papers. Just interesting that the opposite end is genomics. Would have figured bioinformatics or another heavy prevalent category (maybe microbiology). Nevertheless, the exciting part is I can now take varying views of the bioRxiv landscape and see what other discovers I can make (new PR with those visuals coming soon). 

>I just wanted to let you know about https://radimrehurek.com/gensim/models/keyedvectors.html#gensim.models.keyedvectors.WordEmbeddingsKeyedVectors.similar_by_vector if you weren't aware of it.

@MarvinT  Thanks for pointing this out. Turns out I'm doing the same thing in my code, but the link also highlights how I could have done everything in one line rather than a couple. Shows that I should be reading the docs more. Also, feel free to take a look at any past or future PRs if you'd like. (no pressure though) Appreciate any and all feedback.

PR #5, opened 2020-04-10
Figure fix

danich1:
Fixed plot titles to be more straight-forward and edited figures that needed to be fixed for the concept paper. You have already seen this code, so review should be fast.


ajlee21 (approved):
Figures look really nice!

The only minor thing of note is I'm not sure if you want to make the "word dim" a variable so that you can use that to auto-create the titles? Or maybe this is just a one-off quick analysis and its not necessary here. Just thought I'd check

danich1:
> The only minor thing of note is I'm not sure if you want to make the "word dim" a variable so that you can use that to auto-create the titles? Or maybe this is just a one-off quick analysis and its not necessary here. Just thought I'd check

Ah the title change for word dim is to make sure people don't confuse the word embedding dimensions with pca dimensions. But in this analysis it is a one-off quick plot, so I don't believe creating a variable will be needed here; however, in general you are correct that a variable should be used when dealing with multiple values.

PR #8, opened 2020-05-13
Refactoring annorxiver

danich1:
This PR is designed to refactor the current code for annorxiver. I reduced redundancy of my code by moving selected functions and classes to an external file within the modules folder. I also updated the pca figures to have a white background instead of a transparent one. 
Should be a quick review as the only new thing is the document_helper.py file.


modules/document_helper.py
```diff
@@ -0,0 +1,122 @@
...
+def dump_article_text(file_path, xpath_str, filter_tags=filter_tag_list, remove_stop_words=True):
+    """
+    This method is designed to extract all text from xml documents. 
+    Every document has specific tags that are striped in order to produce
+    clean text output for downstream processing
+    
+    Keyword arguments:
+        file_path - the file path for xml document
+        xpath_str - the xpath string to extract tags from the xml document
+        filter_tag_list - the list of tags to strip from the xml document
```
ajlee21, line 29: Just curious, if you can give me an example of a tag. I'm thinking of part-of-speech tags, which I don't think should differ between documents.

modules/document_helper.py
```diff
@@ -0,0 +1,122 @@
...
+    This method is designed to construct document vectors for a given xml document. 
+    Every document has specific tags that are striped in order to have accurate embeddings
+    
+    Keyword arguments:
+        model - the model to extract word vectors from
+        xpath_str - the xpath string to extract tags from the xml document
+        filter_tag_list - the list of tags to strip from the xml document
+    """
+    
+    parser = ET.XMLParser(encoding='UTF-8', recover=True)
```
ajlee21: I'm a little confused. I assumed `dump_article_text` was parsing your article=document and then `generate_doc_vector` took that parsed document (i.e list of words in the document) and embedded it into a vector (i.e. number of occurrences of words)?

Does it make sense to have a separate parser function if being used in two different functions?

ajlee21 (approved):
Really nice refactor. Just a couple questions to clarify what is happening, more for my own knowledge

modules/document_helper.py
danich1: Ah right. In this context tag refers to xml tags, which are akin to HTML website tags. [More Info Here](https://www.w3schools.com/xml/xml_syntax.asp). An example tag would be: 
```xml
<p>
  <italic>I</italic> like machine learning.
</p>
```

modules/document_helper.py
danich1: Good catch. Since parser is being used in both functions it would make more sense to have it be a global variable. I'll make that change.

You are also correct. `dump_article_text` takes in a document and outputs extracted text.  `generate_doc_vector` takes in a document and outputs a vector corresponding to an average of all words within a document.

PR #9, opened 2020-05-20
Comparing Corpora on a Global and Local Scale

danich1:
Have quite a bit going on here, but that is the fun part of code review. This PR is designed to calculate comparison statistics between two corpora (in my case bioRxiv and Pubmed Central).
The money making notebook is `biorxiv_pmc_corpora_comparison.ipynb` which contains the comparison itself.

The other notebooks are set up notebooks:
1. I mapped preprint doi to their published doi versions `map_preprints_to_journals.ipynb`
2. I counted the word frequencies after changing every word to their stem verlsion (lemmatization in NLP language) `biorxiv_pmc_word_frequency_counts.ipynb`

Lastly I let a couple of refactoring code trickle in here. Feel free to ignore that. Should have been in #8, but most have slipped my mind when I closed that PR.


biorxiv/corpora_comparison/biorxiv_pmc_corpora_comparison.py
```diff
@@ -0,0 +1,445 @@
...
+print("Number of words in Pubmed Central but not in biorxiv:")
+pmc_difference = set(list(pmc_corpus_count.keys())) - set(list(biorxiv_corpus_count.keys()))
+print(len(pmc_difference))
+
+
+# In[12]:
+
+
+[
+    pmc_difference.pop()
```
ben-heil, line 223: It may be safer to use `pmc_difference[i]` here in case you decide to do something with `pmc_difference` later in the notebook at some point. If you don't plan on making further changes to the notebook it probably doesn't matter though

biorxiv/corpora_comparison/biorxiv_pmc_corpora_comparison.py
```diff
@@ -0,0 +1,445 @@
...
+
+
+# In[6]:
+
+
+pickle.dump(biorxiv_corpus_count, open("output/biorxiv_total_count.pkl", "wb"))
+pickle.dump(pmc_corpus_count, open("output/pmc_total_count.pkl", "wb"))
+
+
+# ## Analysis without Stop Words
```
ben-heil, line 160: You may want to write a sentence or two about what the analysis actually is

biorxiv/corpora_comparison/biorxiv_pmc_corpora_comparison.py
```diff
@@ -0,0 +1,445 @@
...
+
+
+# In[15]:
+
+
+(
+    total_word_stats_df
+    .sort_values("log_likelihood", ascending=False)
+    .head(20)
+)
```
ben-heil, line 271: All the p_values in this cell and the previous cell are the same. I assume that they're just the log of your epsilon value, but it may be worth checking.

biorxiv/corpora_comparison/biorxiv_pmc_corpora_comparison.py
```diff
@@ -0,0 +1,445 @@
...
+
+# In[14]:
+
+
+total_word_stats_df = pd.DataFrame.from_records(data)
+total_word_stats_df.to_csv(
+    "output/full_corpus_comparison_stats.tsv", 
+    sep="\t", index=False
+)
+total_word_stats_df.head()
```
ben-heil, line 261: ~25% of the memory in this dataframe is taken up by `biorxiv_total` and `pmc_total`. You may want to convert them to variables outside the dataframe since they should have the same value for all rows

ben-heil: It's interesting that "et al." occurs way more frequently in Biorxiv. Do you know whether that's a difference in journal citation styles as a result of some fields publishing more in Biorxiv than others?

biorxiv/corpora_comparison/biorxiv_pmc_corpora_comparison.py
```diff
@@ -0,0 +1,445 @@
...
+
+# In[31]:
+
+
+(
+    published_comparison_stats_df
+    .sort_values("odds_ratio", ascending=False)
+    .head(20)
+)
+
```
ben-heil, line 477: Very cool! Do you have any takeaways from these tables? Are the words that are more common in published manuscripts added due to peer review? Or editing? Or changes in style between the draft and the final version?

You may want to add a few sentences of discussion to the bottom of the notebook with the big points of the two analyses

biorxiv/corpora_comparison/biorxiv_pmc_word_frequency_counts.py
```diff
@@ -0,0 +1,156 @@
...
+get_ipython().run_line_magic('autoreload', '2')
+
+from collections import Counter, defaultdict
+import csv
+from pathlib import Path
+import pickle
+import string
+import sys
+
+sys.path.append("../../modules/")
```
ben-heil, line 19: Sorry to make repo structure comments one PR after a refactor, but if you set up your repo as a python module then you can use standard import syntax for importing `document_helper` without having to tinker with the `PYTHONPATH`

biorxiv/corpora_comparison/biorxiv_pmc_word_frequency_counts.py
```diff
@@ -0,0 +1,156 @@
...
+        writer.writeheader()
+        
+        lemma_freq = Counter(
+            list(
+                map(
+                    lambda x: (
+                        x.lemma_.lower() 
+                        if x.lemma_.lower() != '-pron-' 
+                        else x.text.lower()
+                    ),
```
ben-heil: Adding a comment for this lambda function would be helpful, but it's probably just because I'm not familiar with `spacy`

biorxiv/corpora_comparison/biorxiv_pmc_word_frequency_counts.py
```diff
@@ -0,0 +1,156 @@
...
+    tokens = [tok for tok in doc if tok.text.lower() not in string.punctuation]
+    with open(f"output/pmc_word_counts/{document[1]}.tsv", "w") as file:
+        writer = csv.DictWriter(file, fieldnames=["lemma", "count"],delimiter="\t")
+        writer.writeheader()
+
+        lemma_freq = Counter(
+            list(
+                map(
+                    lambda x: (
+                        x.lemma_.lower()
```
ben-heil: Since you use this twice, it should probably be a full function rather than a lambda function

biorxiv/journal_tracker/map_preprints_to_journals.py
```diff
@@ -0,0 +1,203 @@
...
+    .doi
+    .unique()
+)
+print(len(dois))
+
+
+# In[4]:
+
+
+cf = Crossref(mailto="dnicholson329@gmail.com")
```
ben-heil: You may want to load your email from a config file or something. If you end up with a lot of people running this code, they may get blocked or rate limited for all using the same email

biorxiv/journal_tracker/map_preprints_to_journals.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+# In[5]:
+
+
+TEN_MINUTES = 600
+
+@sleep_and_retry
+@limits(calls=10, period=TEN_MINUTES)
```
ben-heil, line 55: Those decorators are so useful, I hadn't seen the `ratelimit` library before

ben-heil (approved):
Looks good!

biorxiv/journal_tracker/map_preprints_to_journals.py
```diff
@@ -0,0 +1,203 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# # Following the Preprint to Published Path
+
+# The goal of this notebook is to map preprint dois to published dois and published dois to Pubmed Central articles.
```
ajlee21, line 6: Silly question, why are you mapping preprint doi to published doi as opposed to just published doi to articles?

biorxiv/journal_tracker/map_preprints_to_journals.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+# In[5]:
+
+
+TEN_MINUTES = 600
+
+@sleep_and_retry
+@limits(calls=10, period=TEN_MINUTES)
+def call_crossref(doi_ids):
```
ajlee21, line 56: Could pull this function out if you wanted, but not a big deal either way

biorxiv/journal_tracker/map_preprints_to_journals.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+journal_mapper_df = pd.DataFrame.from_records(journal_mapper_records)
+journal_mapper_df.head()
+
+
+# In[12]:
+
+
+final_df = (
```
ajlee21, line 156: Just curious, what are the values in the "header"? Sounds like these are the status of the articles..?

biorxiv/journal_tracker/map_preprints_to_journals.py
```diff
@@ -0,0 +1,203 @@
...
+)
+
+
+# In[14]:
+
+
+batch_limit = 100
+journal_mapper_records = []
+
+for batch in tqdm.tqdm(range(0, len(journal_dois), batch_limit)):
```
ajlee21, line 131: Could pull out this parser if you find you're using it multiple times. 

Curious, is there a reason to read in in batches? Memory?

biorxiv/journal_tracker/map_preprints_to_journals.py
```diff
@@ -0,0 +1,203 @@
...
+
+# # Map preprint DOIs to Published DOIs
+
+# In[8]:
+
+
+batch_limit = 100
+doi_mapper_records = []
+
+for batch in tqdm.tqdm(range(0, len(dois), batch_limit)):
```
ajlee21, line 81: How is this mapping from preprint doi to pub doi working? Is this in the json somewhere?

biorxiv/journal_tracker/map_preprints_to_journals.py
ajlee21: I think I'm behind on what your project is. I thought that you were looking into categories of preprints..so I'm not sure why you're comparing things here. Perhaps this will get added to the readme later though

ajlee21 (approved):
Interesting! So is the idea that there are different distributions of words that make up those preprints that do not make it to publication vs those that are published? Is the takeaway that the mention of figures and supplemental material is important to publish..?

biorxiv/corpora_comparison/biorxiv_pmc_corpora_comparison.py
danich1: Ah good point. This line is a sanity check to show that words that appear in PMC and not biorxiv are mainly parse errors. Don't plan to do anything else with the variable.

biorxiv/corpora_comparison/biorxiv_pmc_corpora_comparison.py
danich1: Yeah. The chi-squared test breaks when numbers reach epic proportions. As a consequence everything becomes "significant", which makes hypothesis testing pretty difficult. This is also why I added log-likelihood and odds ratio calculations.

biorxiv/corpora_comparison/biorxiv_pmc_corpora_comparison.py
danich1: That is also my hypothesis as to why et al. appears so frequently in bioRxiv rather than PMC. I think journals have their own special way of formatting citations, which diverges from the common preprint citation style which uses <author> et al. Interesting discovery to say the least.

biorxiv/journal_tracker/map_preprints_to_journals.py
danich1: Fantastic catch. Yeah my email should not be there. I definitely __don't__ want people running my code and making me be responsible for their foolishness.

biorxiv/journal_tracker/map_preprints_to_journals.py
danich1: > Silly question, why are you mapping preprint doi to published doi as opposed to just published doi to articles?

The bioRxiv dump doesn't include published version dois. 🤷  For future analyses I need to know what preprints have been published and where; hence the necessary doi mapping step.

>I think I'm behind on what your project is. I thought that you were looking into categories of preprints..so I'm not sure why you're comparing things here. Perhaps this will get added to the readme later though

Whoops forgot to include that description. So this analysis was prompted after my committee meeting. Long story short I needed to take one step back and answer a fundamental question: "What is the difference between the bioRxiv corpus and Pubmed Central corpus?".  To answer this question I needed to compare word frequencies between corpora on a global scale (bioRxiv v PMC) and local scale (preprint version vs published versions).

biorxiv/journal_tracker/map_preprints_to_journals.py
danich1: Yes I could; however, this is the only notebook where I am using this function. If I use it anywhere else in the future I will move it to my modules folder.

biorxiv/journal_tracker/map_preprints_to_journals.py
danich1: The header field describes what type of preprint is being submitted. Is the preprint a new finding? Is the preprint highlighting confirmatory results or even contradictory results? 
Basically most preprints are new findings as expected, but there are confirmatory and contradictory results.

biorxiv/journal_tracker/map_preprints_to_journals.py
danich1: Luckily this is the only time I am using this parser. 

>Curious, is there a reason to read in in batches? Memory?

The main reason here is to throttle my api calls and avoid being blocked/yelled at by the crossref developers. Plus if anything happened to my parser I could track progress and restart midway rather than from the beginning.

biorxiv/journal_tracker/map_preprints_to_journals.py
danich1: > How is this mapping from preprint doi to pub doi working? Is this in the json somewhere?

It works like this:
1. Take a preprint doi and ask cross ref is there a published doi that matches this preprint?
2. If yes capture the published doi version otherwise skip the preprint doi
3. The the next step is to cycle through the published doi and ask the following question
4. What is the journal that maps to this doi? If not found put a place holder and make @danich1  manually fill it in. Only had a couple that I needed to fill in.

The call_crossref makes an api call to the crossref server and the responses variable is a list of json objects that gets returned.

danich1:
>Interesting! So is the idea that there are different distributions of words that make up those preprints that do not make it to publication vs those that are published?

Close for the first part. I'm comparing preprints to their published versions so I'm ignore preprints that haven't been published.

>Is the takeaway that the mention of figures and supplemental material is important to publish..?

Yup. That is exactly right. As I mentioned in one of my responses: the main goal here is to literally compare words and their frequencies between corpora. A big finding here is that focus shifts to adding figures and supplementary data to articles before they are published. Plus, citation style changes as papers are sent to journals.

biorxiv/corpora_comparison/biorxiv_pmc_word_frequency_counts.py
danich1: Noted! I'll defer this point for another pull request as this one has gotten a bit big on changes.

PR #10, opened 2020-05-27
Journal Recommendation Experiment

danich1:
This PR is not ready for review yet, but the main idea behind this is to predict which journals are the most appropriate for a given preprint. Performance has been modest when using full 300 dimensions, but will need to check performance with lower number of dimensions before review.


danich1:
Ready for review. @ajlee21 and @ben-heil

annorxiver_modules/journal_rec_helper.py
```diff
@@ -0,0 +1,197 @@
...
+    train_centroid_df = (
+        training_data
+        .groupby("journal")
+        .agg("mean")
+        .reset_index()
+    )
+            
+    X_train_centroid = (
+        train_centroid_df
+        .drop("journal", axis=1)
```
ben-heil, line 164: I may be wrong because I'm not sure exactly what the columns of this dataframe are, but is this logic behind this selection "Pick the column that isn't 'journal'"? If so, wouldn't it be more explicit to select by column like you do in `Y_train_centroid`?

biorxiv/journal_tracker/journal_recommendation_experiment.py
```diff
@@ -0,0 +1,354 @@
...
+    .reset_index(drop=True)
+)
+biorxiv_journal_df.head()
+
+
+# In[3]:
+
+
+# Count number of Non-NaN elements
+print(f"Number of Non-NaN entries: {biorxiv_journal_df.pmcid.count()}")
```
ben-heil, line 55: Are Non-NaN entries the ones that have been published?

biorxiv/journal_tracker/journal_recommendation_experiment.py
```diff
@@ -0,0 +1,354 @@
...
+print(f"Total number of entries: {biorxiv_journal_df.shape[0]}")
+print(f"Percent Covered: {biorxiv_journal_df.pmcid.count()/biorxiv_journal_df.shape[0]:.2f}")
```
ben-heil: This is a little misleading because it's not a percentage

biorxiv/journal_tracker/journal_recommendation_experiment.py
```diff
@@ -0,0 +1,354 @@
...
+pmc_articles_df = pmc_articles_df.query(f"journal in {journals[journals > 100].index.tolist()}")
+print(pmc_articles_df.shape)
+pmc_articles_df.head()
+
+
+# In[8]:
+
+
+pmc_embedding_dict = {
+    int(path.stem[-7:-4]):pd.read_csv(
```
ben-heil: This is kind of fragile, do you know that you'll always have exactly three digit identifiers that are always in the -7 to -4 position? Might it make more sense to just use a delimiter in the file names?

biorxiv/journal_tracker/journal_recommendation_experiment.py
```diff
@@ -0,0 +1,354 @@
...
+# In[30]:
+
+
+print(f"{np.sum(accs)} out of {len(accs)}")
+print(f"{np.mean(accs)}% correct")
+
+
+# Conclusions for this notebook:
+# 1. Mega-journals cover a wide range of research topics.
+# 2. The correct journal only appears in the top ten about 37-39 percent of the time.
```
ben-heil, line 458: I'm not sure whether your goal is to max out prediction accuracy or just use a simple model to show that the embeddings are predictive of which journal the articles end up in.

If you're looking to get as accurate as possible, manhattan distance or a fractional distance may outperform euclidean distance [due to high dimensionality](https://bib.dbvis.de/uploadedFiles/155.pdf).

ben-heil (approved):
Looks good! I left a couple style comments and a potential modeling change that you can take or leave

biorxiv/journal_tracker/journal_recommendation_experiment.py
ajlee21: Just to clarify, 24% of biorxiv pre-prints have been published (non-NaN published doi) and have full text available (non-NaN pmcid). These make up your golden set for which you have document vectors?

biorxiv/journal_tracker/journal_recommendation_experiment.py
```diff
@@ -0,0 +1,354 @@
...
+
+
+# In[4]:
+
+
+golden_set_df = biorxiv_journal_df.query("pmcid.notnull()")
+golden_set_df.head()
+
+
+# # Load Pubmed Central Document Vectors
```
ajlee21, line 67: These are already published documents, compared to the biorxiv that is sometimes published?

biorxiv/journal_tracker/journal_recommendation_experiment.py
```diff
@@ -0,0 +1,354 @@
...
+    )
+    for path in Path("../../pmc/word_vector_experiment/output/").rglob("*tsv.xz")
+}
+pmc_embedding_dict[300].head()
+
+
+# In[9]:
+
+
+full_training_dataset = {
```
ajlee21, line 121: So you're training set is composed of all published articles EXCEPT those that are currently matched to preprints?

biorxiv/journal_tracker/journal_recommendation_experiment.py
```diff
@@ -0,0 +1,354 @@
...
+
+
+# ## Random Journal Prediction
+
+# The central idea here is to answer the question what is the accuracy when journals are recommended at random?
+
+# In[12]:
+
+
+model = DummyClassifier(strategy='uniform')
```
ajlee21, line 167: You're random model is assigning journal recommendations with a uniform distribution. i.e. article A is equally likely (1/N) to be assigned to any of the  N categories of journals

biorxiv/journal_tracker/journal_recommendation_experiment.py
```diff
@@ -0,0 +1,354 @@
...
+_ = cross_validation(
+    knn_model, full_training_dataset[300], 
+    knn_evaluate, cv=10,
+    random_state=100
+)
+
+
+# ## KNearestNeighbors Paper by Paper Comparison Subsampled
+
+# The first idea for a classifier is to compare which papers are similar to other papers. Due to the overflow of PLOS One papers I sub-sampled each journal to have only 100 papers for representation. Then trained a KNearestNeighbors to determine how often does the correct journal appear in the top ten neighbors as well as top twenty neighbors.
```
ajlee21, line 196: Here you're predicting which a set of possible journal to recommend to article A based on which journals similar articles were published in? And the metric you're using is if the actual journal is in the set of possible journals?

biorxiv/journal_tracker/journal_recommendation_experiment.py
```diff
@@ -0,0 +1,354 @@
...
+
+
+# ## KNearestNeighbors Paper by Paper Comparison - Full Dataset
+
+# Assuming I didn't take mega-journal influence into account, what would the initial recommendation accuracy be?
+
+# In[ ]:
+
+
+_ = cross_validation(
```
ajlee21, line 187: What is your similarity metric here? Are two articles considered similar if their 300-dim vectors are similar? Do you know if articles cluster in this space? Is it better to use some other smaller representation?

biorxiv/journal_tracker/journal_recommendation_experiment.py
```diff
@@ -0,0 +1,354 @@
...
+     )
+     for data_idx, prediction_row in enumerate(predictions)
+]
+
+
+# In[21]:
+
+
+print(f"{np.sum(accs)} out of {len(accs)}")
+print(f"{np.mean(accs)}% correct")
```
ajlee21: I'm outside of my field on this, do you think 40% accuracy pretty good in this case, given the 9K journals you're using

ajlee21 (approved):
Nice analysis! Just some comments to try to better understand what you're trying to do here

annorxiver_modules/journal_rec_helper.py
danich1: >I may be wrong because I'm not sure exactly what the columns of this dataframe are, but is this logic behind this selection "Pick the column that isn't 'journal'"?

The dataframe contains the column that specifies the journal each article belongs and the document vectors of 300 dimensions. 

>If so, wouldn't it be more explicit to select by column like you do in Y_train_centroid?

To save time it is easier to just drop the journal column rather than select all the features.

biorxiv/journal_tracker/journal_recommendation_experiment.py
danich1: >Are Non-NaN entries the ones that have been published?

Correct. The Non-NaN entries are preprints where the published version has been found.

>Just to clarify, 24% of biorxiv pre-prints have been published (non-NaN published doi) and have full text available (non-NaN pmcid). These make up your golden set for which you have document vectors?

Correct.

biorxiv/journal_tracker/journal_recommendation_experiment.py
danich1: >I'm not sure whether your goal is to max out prediction accuracy or just use a simple model to show that the embeddings are predictive of which journal the articles end up in.

The goal is to show that embeddings are predictive of which journal the articles end up; however, I want to be as accurate as possible. Long sentence converted to a short one: the goal is to do both.

>If you're looking to get as accurate as possible, manhattan distance or a fractional distance may outperform euclidean distance due to high dimensionality.

Thanks will see how performance changes.

biorxiv/journal_tracker/journal_recommendation_experiment.py
danich1: >These are already published documents, compared to the biorxiv that is sometimes published?

Yup. Everything in Pubmed Central is published.

biorxiv/journal_tracker/journal_recommendation_experiment.py
danich1: > So you're training set is composed of all published articles EXCEPT those that are currently matched to preprints?

Correct. This is to prevent any leakage when evaluating the golden set.

biorxiv/journal_tracker/journal_recommendation_experiment.py
danich1: >You're random model is assigning journal recommendations with a uniform distribution. i.e. article A is equally likely (1/N) to be assigned to any of the N categories of journals

Yup. You are correct. This is just to show what performance would be if all journals were equally likely to be selected.

biorxiv/journal_tracker/journal_recommendation_experiment.py
danich1: > Here you're predicting which a set of possible journal to recommend to article A based on which journals similar articles were published in?

Correct. I'm finding articles that are as close to a given article and noting which journal these articles were published in. 

>And the metric you're using is if the actual journal is in the set of possible journals?

Correct. I'm evaluating my model by answer the question: Does the correct journal appear in the top 10 most closest neighbors to my query? If yes then positive and negative otherwise

biorxiv/journal_tracker/journal_recommendation_experiment.py
danich1: >What is your similarity metric here? Are two articles considered similar if their 300-dim vectors are similar?

Euclidean distance. Basically it is the L2 norm of two document vectors: 

>Do you know if articles cluster in this space?

Articles definitely cluster as I manually inspected a few predictions when running the different models. Now how neat these clusters are I do not know; however, based on my intuition I'd assume a 2D projection would result in a large blob with no easy distinction between categories.

> Is it better to use some other smaller representation?

Good question I'm not sure. It could be interesting to see if a smaller representation could help. I'll see if that helps anything. Thanks for the suggestion.

biorxiv/journal_tracker/journal_recommendation_experiment.py
danich1: > I'm outside of my field on this, do you think 40% accuracy pretty good in this case, given the 9K journals you're using

Well it is subsampled to 2K as I cut off journals with low representation (<100). Since this is the first time this analysis has been done, 40% is acceptable. I'm sure as soon as I release this to the public 40% will shoot up to possible >60% in a few years. Just a hard problem as paper content alone doesn't provide a robust way to predict journal endpoints. Predicting peer review is a challenge within itself.

danich1:
Turns out Euclidean distance performs slightly better than Manhattan distance. On the magnitude of a fraction of a percentage. Stayed away from fractional distance because it has been refuted in [this paper](https://arxiv.org/pdf/2004.14230.pdf). Gonna merge these results and another PR may pop up with a figure that summarizes this work.

PR #11, opened 2020-05-27
Figure for Corpora Comparison

danich1:
This PR contains a figure I will be using in my annorxiver manuscript. The central idea is to create a figure that:

1. Conveys which repository a word indicates.
2. How frequent a word is relative to other words.
3. How specific a word is to the repository in question.

My initial answer for this is a treemap. Let me know what you guys think/if you think of something better.


danich1:
Hold off on review. Chatted with @cgreene and I have a new figure coming. Will ping you both when ready.

danich1:
Thanks for waiting! Feel free to review!

biorxiv/corpora_comparison/biorxiv_pmc_figure_generator.py
```diff
@@ -0,0 +1,165 @@
...
+    .sort_values("odds_ratio", ascending=False)
+)
+full_text_comparison
+
+
+# In[4]:
+
+
+subset = 20
+plot_df = (
```
ajlee21, line 53: Can you remind me what _total counts are? Are these the total number of occurrences of all words, thats why they are the same for each word. So your odds ratio is the enrichment of the occurrences of a given word in that repo compared to all words occurring?

biorxiv/corpora_comparison/biorxiv_pmc_figure_generator.py
```diff
@@ -0,0 +1,165 @@
...
+        'yanchor': 'top'}
+)
+fig.update_layout(uniformtext=dict(minsize=10))
+fig.show()
+
+
+# In[11]:
+
+
+fig.write_image("output/figures/preprint_published_comparison.svg", scale=2.5)
```
ajlee21: Do you want to remove the degree symbol and single letters for this analysis?

Also for preprint vs published, are these preprints that have NOT been published? So is the published a subset of the preprints? I assume not given the strong difference in words but just wanted to check.

biorxiv/corpora_comparison/biorxiv_pmc_figure_generator.py
```diff
@@ -0,0 +1,165 @@
...
+        labels = ["preprint","published"] + plot_df.label.tolist(),
+        parents = ["", ""] + plot_df.repository.tolist(),
+        values = [0,0] + plot_df.odds_ratio.tolist(),
+        textinfo = "label",
+        textfont = {
+            'size':15
+        }
+    )
+)
+fig.update_layout(
```
ajlee21, line 317: I'm not sure the best way to visualize this information. My instinct would just be to have two ranked lists of words because it would be straightforward to compare. 

The only issues I had with this visualization is that the smaller words I ended up ignoring when i looked at it just because their font was so small, though i understand this was to show that it was less specific. I think I personally had a little trouble going back and forth between the two plots to compare the words and the words were scattered around so I would loose my place a little when I'd glace back.

In short I'm not sure of another option other than ranked lists right now and i I think the way you have it is a good start.

ajlee21 (approved):
Nice PR, just a few questions

biorxiv/corpora_comparison/biorxiv_pmc_figure_generator.py
danich1: >Are these the total number of occurrences of all words, that's why they are the same for each word. 

Correct. These are the total number of words found within a given corus.

>So your odds ratio is the enrichment of the occurrences of a given word in that repo compared to all words occurring?

Almost the odds ratio is measuring the association of the chance that a given word appears in one corpus rather than another. For example a ratio of >1 would indicate a word appears more often in bioRxiv rather than in PubMed Central. Likewise ratio <1 would indicate that a word appears more in PubMed Central than bioRxiv

biorxiv/corpora_comparison/biorxiv_pmc_figure_generator.py
danich1: > Do you want to remove the degree symbol and single letters for this analysis?

I thought about removing the single letters and symbols, but ultimately decided to keep them in the analysis as they pertain to math symbols and measurements. Good to note for future analysis of this dataset!

> Also for preprint vs published, are these preprints that have NOT been published? So is the published a subset of the preprints?

No. These are only the preprints that have gone through the publishing process. The idea here is to see how a research article changes when it has been accepted by a journal.

ajlee21: Ah..so does that mean the articles included in the preprint and published should be the same, but the text content might be different and thats what you're comparing?

biorxiv/corpora_comparison/biorxiv_pmc_figure_generator.py
ajlee21: That makes sense i see, thanks

biorxiv/corpora_comparison/biorxiv_pmc_figure_generator.py
danich1: >Ah..so does that mean the articles included in the preprint and published should be the same, but the text content might be different and thats what you're comparing?

Bingo! That is correct. The expectation here is that text content should change as reviewers and editors suggest revisions towards an article.

biorxiv/corpora_comparison/biorxiv_pmc_figure_generator.py
danich1: Which figure did you see? Was it the blue and green treemap? If so that is the wrong figure. You were suppose to see a point plot with a barchart. I updated the graph. Please take a look at this one.

danich1:
@ajlee21 and @ben-heil: I updated the figure, so hopefully that is resolved. Let me know what you think on the current version.

biorxiv/corpora_comparison/biorxiv_pmc_figure_generator.py
ajlee21: Yup looked at the wrong plot!

These bar plots look a lot better in terms of comparing the data. Its interesting that the terms with odd ratio >1 reflect very slight increases in the preprints, but for terms with odd ratio <1 the difference in counts is much larger in the published papers.  Not sure if it makes more sense to sort by the change in reference to 1 for this

biorxiv/corpora_comparison/biorxiv_pmc_figure_generator.py
```diff
@@ -0,0 +1,486 @@
...
+
+
+plot_df = (
+    full_text_comparison_old.head(subset).append(full_text_comparison_old.tail(subset))
+    .assign(odds_ratio=lambda x: x.odds_ratio.apply(lambda ratio: ratio if ratio > 1 else 1/ratio))
+    .assign(label=lambda x: x.apply(
+        lambda row: (
+            f"<b>{row.lemma}</b><br />bioRxiv:{row.biorxiv_count:,}"
+            f"<br />PMC:{row.pmc_count:,}<br />"
+            f"Odds Ratio:{row.odds_ratio:.2f}"
```
ben-heil, line 60: Alex already addressed this, but you may want to add expand the axis label here to be "bioRxiv vs PMC Odds Ratio" for the sake of clarity

biorxiv/corpora_comparison/biorxiv_pmc_figure_generator.py
```diff
@@ -0,0 +1,486 @@
...
+        textinfo = "label",
+        textfont = {'size':17}
+    )
+)
+fig.update_layout(
+    width=1280, 
+    height=720,
+    treemapcolorway=["#a6cee3", "#b2df8a"],
+    title={
+        'text':f"Comparison of Top {plot_df.shape[0]} Words between bioRxiv and Pubmed Central (PMC)",
```
ben-heil: Top 40 by what metric? Most common? Most specific to that venue?

biorxiv/corpora_comparison/biorxiv_pmc_figure_generator.py
```diff
@@ -0,0 +1,486 @@
...
+        'y':0.9,
+        'x':0.5,
+        'xanchor': 'center',
+        'yanchor': 'top'}
+)
+fig.update_layout(uniformtext=dict(minsize=10))
+fig.show()
+
```
ben-heil, line 95: This interactive tree plot is so cool!

biorxiv/corpora_comparison/biorxiv_pmc_figure_generator.py
```diff
@@ -0,0 +1,486 @@
...
+        figure_size=(13.653333333333334, 10.24),
+        axis_text_y=p9.element_text(family='DejaVu Sans', size=12),
+        panel_grid_minor=p9.element_blank(),
+        axis_title=p9.element_text(size=15),
+        axis_text_x=p9.element_text(size=11, weight="bold")
+    )
+    + p9.labs(x=None)
+)
+g.save("output/svg_files/preprint_published_frequency_odds.svg", dpi=75)
+print(g)
```
ben-heil, line 407: It's interesting that the long tail in this lineplot is on the <1 side while it was on the >1 side in the PMC vs BioRxiv line plot. Is there a reason for that? Or just random outliers?

biorxiv/corpora_comparison/biorxiv_pmc_figure_generator.py
```diff
@@ -0,0 +1,486 @@
...
+# In[23]:
+
+
+g = (
+    p9.ggplot(count_plot_df, p9.aes(x="lemma", y="count"))
+    + p9.geom_col(position=p9.position_dodge(width=0.5))
+    + p9.coord_flip()
+    + p9.facet_wrap("category")
+    + p9.scale_x_discrete(limits=plot_df.sort_values("odds_ratio", ascending=True).lemma.tolist())
+    + p9.scale_y_continuous(labels=custom_format('{:,g}'))
```
ben-heil, line 445: It makes sense to sort by odds ratio here, but the order of the image will be confusing unless it ends up next to the line plots in the publication

ben-heil (approved):
Awesome! Sorry it took me awhile to see this PR, I hadn't checked my github notifications in a few days. Added a couple requests of axis label clarifications and things

biorxiv/corpora_comparison/biorxiv_pmc_figure_generator.py
danich1: >It's interesting that the long tail in this lineplot is on the <1 side while it was on the >1 side in the PMC vs BioRxiv line plot. Is there a reason for that? Or just random outliers?

Good catch. Honestly, I think it is just a coincidence. Words are sorted by their frequency, which just happens to fall into this pattern.

biorxiv/corpora_comparison/biorxiv_pmc_figure_generator.py
danich1: >It makes sense to sort by odds ratio here, but the order of the image will be confusing unless it ends up next to the line plots in the publication

Actually the line plots and bar charts are suppose to be aligned together in one figure. I don't know what is going on with my notebook versioning, but I'll update the notebook to clear confusion.

PR #12, opened 2020-06-30
Fill in Missing Biorxiv Links

danich1:
Once a preprint is published, bioRxiv links their preprints to published versions. Problem is that published versions can look remarkably different, which causes issues for establishing preprint-published links. This PR contains experimental results that aim to see if missing links can be filled in via paper embedding distances. 

To speed up review here is a listing of files ranked in order of review preference. Bronze can be ignored.

Golden Files:
biorxiv/article_distances/biorxiv_article_distances.py <- script
biorxiv/article_distances/output/potential_biorxiv_pmc_links.tsv <- the potential links
biorxiv/article_distances/output/biorxiv_article_distance.png <- distribution of articles distances (pre vs publish means preprint compared to its published version, pre vs random indicates a preprint compared to a randomly chosen paper from the same published journal)

Silver Files:
biorxiv/journal_tracker/map_preprints_to_journals.py  <- updated script for filling in missing PMCIDs

Bronze Files:
biorxiv/corpora_comparison/biorxiv_pmc_figure_generator.py <- This is a figure fix for another result.


biorxiv/article_distances/biorxiv_article_distances.py
```diff
@@ -0,0 +1,499 @@
...
+# In[21]:
+
+
+g = (
+    p9.ggplot(final_df)
+    + p9.aes(x="label",y="distance")
+    + p9.geom_boxplot()
+    + p9.coord_flip()
+    + p9.theme_seaborn()
+)
```
ben-heil, line 272: This plot might be clearer as a violin plot or a jitter plot

biorxiv/article_distances/biorxiv_article_distances.py
```diff
@@ -0,0 +1,499 @@
...
+
+# In[24]:
+
+
+cutoff_score = (
+    final_df
+    .query("label=='pre_vs_random'")
+    .distance
+    .min()
+)
```
ben-heil, line 429: If you use regularized logistic regression on the differences between the articles you may be able to get more separation between `pre_vs_random` and `pre_vs_published`. Right now the model is operating under the assumption that all embedded dimensions are equally important for determining whether two articles are the same, which may not be true.

ben-heil (approved):
Looks good! I left some comments on potential alternative visualizatios/approaches to differentiating between articles

biorxiv/article_distances/biorxiv_article_distances.py
```diff
@@ -0,0 +1,499 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# # Find published articles missing from bioRxiv
```
ajlee21, line 4: Just to clarify you are trying to link preprint dois with published doi. What is the current way they are being linked? And so there is a problem with some links not existing that should?

biorxiv/article_distances/biorxiv_article_distances.py
```diff
@@ -0,0 +1,499 @@
...
+    })
+    .reset_index(drop=True)
+)
+biorxiv_journal_df.head()
+
+
+# In[3]:
+
+
+biorxiv_embed_df = (
```
ajlee21, line 49: What is the difference between biorxiv_journal_mapped_df and boirxiv_embed_df?

biorxiv/article_distances/biorxiv_article_distances.py
```diff
@@ -0,0 +1,499 @@
...
+    .sort_values("document", ascending=True)
+    .set_index("document")
+)
+PMC_pubished.head()
+
+
+# In[10]:
+
+
+article_distances = cdist(
```
ajlee21, line 145: This is the distance between all preprints vs all published?

ajlee21: Based on the plot below it looks like this is 1-1 distance between preprint and its associated publication

biorxiv/article_distances/biorxiv_article_distances.py
```diff
@@ -0,0 +1,499 @@
...
+g = (
+    p9.ggplot(final_df)
+    + p9.aes(x="label",y="distance")
+    + p9.geom_boxplot()
+    + p9.coord_flip()
+    + p9.theme_seaborn()
+)
+g.save("output/biorxiv_article_distance.svg", dpi=500)
+g.save("output/biorxiv_article_distance.png", dpi=500)
+print(g)
```
ajlee21, line 275: Looking at this figure, would you intrepret this as dist(preprint, some other article in the same journal ) >> dist(preprint, associated publication), which is what you'd expect. But that the distance of the preprint vs random article is close enough that you could use this to link preprint with a publication?

Are there that many articles with missing links?

biorxiv/article_distances/biorxiv_article_distances.py
ajlee21: Along similar lines, I was going to ask if you tried with fewer dimensions since 300 might be hitting the curse of dimensionality where distances are equal between points.

ajlee21 (approved):
Looks good to me, just some comments to clarify some of the logic for myself to better understand what you're doing.

biorxiv/article_distances/biorxiv_article_distances.py
danich1: >Just to clarify you are trying to link preprint dois with published doi.

Correct. I'm trying to link preprints with their missing publish dois

>What is the current way they are being linked?

Great question. I don't know how biorxiv currently links preprints with published dois. I'd speculate they have their own custom search procedure.

>And so there is a problem with some links not existing that should?

Correct. There are preprints that are missing their published version's doi. If I can establish these links it is another bonus for the manuscript as well as gives biorxiv another way to keep track of preprints.

biorxiv/article_distances/biorxiv_article_distances.py
danich1: >What is the difference between biorxiv_journal_mapped_df and boirxiv_embed_df?

biorxiv_journal_mapped_df - dataframe with biorxiv preprints linked with their published versions
biorxiv_embed_df - dataframe that contains document embeddings for preprints

biorxiv/article_distances/biorxiv_article_distances.py
danich1: > This is the distance between all preprints vs all published?
> Based on the plot below it looks like this is 1-1 distance between preprint and its associated publication

Correct. This is the distance between preprint embeddings and their published counter part embeddings. I'm forcing it to be a 1-1 mapping in this case.

biorxiv/article_distances/biorxiv_article_distances.py
danich1: >Looking at this figure, would you interpret this as dist(preprint, some other article in the same journal ) >> dist(preprint, associated publication), which is what you'd expect. But that the distance of the preprint vs random article is close enough that you could use this to link preprint with a publication?

Correct. The central idea here is to see if there is a difference between preprints to their published version and preprints with a random articles from the same journal. Given this difference, I can devise a cutoff score that would state that this preprint and this published article should have a link because it is the same paper.

>Are there that many articles with missing links?

There are about 39k articles that are missing links, which is a bit high. I could potentially drop that number down by approximately ten percent. Ideally, if more articles were open access I could decrease it by a lot more than ten percent.

PR #13, opened 2020-07-13
Ci intervals for PCA Cossim Plots

danich1:
This PR contains point plots for 95% confidence intervals of the cosine similarity scores between article categories and pca components (PC1 and PC2).

 These intervals were generated via the following algorithm:
1. Group document embeddings by category
2. Randomly sample a document from each category
3. Calculate the similarity scores between the documents and the first two principal components
4. Repeat the above steps for 10000 iterations.
5. Finally take the 25th percentile and the 97.5th percentile to make up the interval


biorxiv/pca_association_experiment/biorxiv_pca_category_bootstrap.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+# In[7]:
+
+
+boostraped_iterations = 10000
+pc_1 = []
+pc_2 = []
+
+for iteration in tqdm_notebook(range(boostraped_iterations)):
```
ben-heil, line 106: You may want to add a comment explaining what this cell is doing, I don't know that I would have been able to follow without the PR description

ajlee21, line 106: This is a lot of pandas magic for me to read! So just to clarfiy, here you're grouping your pca-embedded docs by category and you are calculating the similarity between two categories? So is cosine comparing all members of one category to another since I don't see an aggregation...?

biorxiv/pca_association_experiment/biorxiv_pca_category_bootstrap.py
```diff
@@ -0,0 +1,230 @@
...
+    pc_2.append(document_distance[:, 1])
+
+pc_1 = np.stack(pc_1)
+pc_2 = np.stack(pc_2)
+
+
+# In[8]:
+
+
+mean_similarity_df = (
```
ajlee21, line 137: Here you have the cosine similarity between all documents against each other regardless of category?

biorxiv/pca_association_experiment/biorxiv_pca_category_bootstrap.py
```diff
@@ -0,0 +1,230 @@
...
+        document = document_categories_df.document.tolist()
+    )
+)
+mean_similarity_df.head()
+
+
+# In[9]:
+
+
+category_sim_df = (
```
ajlee21, line 161: Here is where you're combining everything, so you are taking the mean similarity of docs within a category and getting the CI for the similarity of docs in the same category (but i thought the iterations above were comparing between categories not within so not sure how this is a CI)

biorxiv/pca_association_experiment/biorxiv_pca_category_bootstrap.py
```diff
@@ -0,0 +1,230 @@
...
+category_sim_df.to_csv(
+    "output/category_cossim_95_ci.tsv", 
+    sep="\t", index=False
+)
+
+
+# In[14]:
+
+
+g = (
```
ajlee21, line 199: So these are showing the variability of docs within a given category. Looks pretty variable..is this meant to give context for your journal predictions? So that you are able to say predict X well because of the variability of documents in category Y?

ajlee21 (approved):
Looks good to me, just some clarifying comments to help be follow along in your logic. Up to you if you want to add additional comments for these

biorxiv/pca_association_experiment/biorxiv_pca_category_bootstrap.py
```diff
@@ -0,0 +1,230 @@
...
+        .apply(lambda x: x.sample(1, random_state=iteration))
+        .reset_index(drop=True)
+        .sort_values("category")
+    )
+    
+    document_distance = (
+        1 - cdist(
+            sampled_df
+            .drop(["document", "category"], axis=1)
+            .values,
```
ben-heil: At least in my opinion indenting the `.drop` and `.values` calls would make this more readable. It took me a little bit to realize that they were modifying `sampled_df`, not parameters themselves

biorxiv/pca_association_experiment/biorxiv_pca_category_bootstrap.py
danich1: So basically the idea for this for loop is to randomly sample one document from each article category. Once sampled, the next step is to calculate the cosine similarity of each document with the first two principal component directions. Save the result and repeat. The idea here is to get a general measure of the variance between documents within assigned categories.

ben-heil (approved):
Looks good, I'm a little unclear on how to interpret the distribution of the cosine similarities of document embedding against the global PCs of the embeddings though. Is the idea to show how unusual a given category is with respect to the 'normal' embedding in that PC? 

If that's the case what does your distribution of categories look like? Is it possible that your PCs fail to represent a normal document because variation could be driven by an idiosyncrasy in a single category or journal?

biorxiv/pca_association_experiment/biorxiv_pca_category_bootstrap.py
danich1: >Here you have the cosine similarity between all documents against each other regardless of category?

I'm calculating the cosine similarity of all documents with the first two principal component directions. Idea here is to get the mean similarity score for each category.

biorxiv/pca_association_experiment/biorxiv_pca_category_bootstrap.py
danich1: > (but i thought the iterations above were comparing between categories not within so not sure how this is a CI)

No the similarity scores are between embeddings and principal components. After all the iterations have been generated, I'm taking the 25th and 97.5th percentiles to be my 95% confidence intervals for each category. This is a brute force way of estimating CI when I don't know the underlying distribution (called Bootstrapping).

biorxiv/pca_association_experiment/biorxiv_pca_category_bootstrap.py
danich1: > is this meant to give context for your journal predictions? So that you are able to say predict X well because of the variability of documents in category Y?

No. This is more along the lines of highlighting how categories are related with a given principal component. The idea here is to give the reader a sense of the type of research is being discussed in various preprints. For example the biochemistry category has a mean at 0 and a large variance for the PC1. PC1 is about molecular biology vs quantitative biology. This means on average biochemistry preprints contain an equal amount of molecular and quantitative experiments; however, a greater number of preprints contain quantitative biology as shown by the positive skew of the variance.

danich1:
> Is the idea to show how unusual a given category is with respect to the 'normal' embedding in that PC?
>If that's the case what does your distribution of categories look like? Is it possible that your PCs fail to represent a normal document because variation could be driven by an idiosyncrasy in a single category or journal?

It's not about a 'normal' document. It's more along the lines of depicting the type of research being conducted within each category. For example, looking at the first PC plot. PC1 is about molecular biology vs quantitative biology so categories that have a high similarity with PC1 contain preprints that mainly discuss quantitative biology experiments and results. Likewise categories that are very dissimilar contain preprints that more focused on molecular biology experiments and results. Good examples for both sides would be bioinformatics (positive) and immunology (negative).

cgreene:
Is this in response to my question on your manuscript? If so, I think I may have been thinking _much_ simpler (just breaking out each category separately and plotting their value in PC1 and also PC2). Doesn't each document have a value in PC1, PC2...? Maybe I am misunderstanding these PCs.

danich1:
> Is this in response to my question on your manuscript? If so, I think I may have been thinking much simpler (just breaking out each category separately and plotting their value in PC1 and also PC2). Doesn't each document have a value in PC1, PC2...? Maybe I am misunderstanding these PCs.

Yeah. This experiment was in response to your comment. I took it in a totally different direction. Whoops. Anyways, yes each document has been projected onto PC1 and PC2. Plotting that would have be a lot simpler.

biorxiv/pca_association_experiment/biorxiv_pca_category_bootstrap.py
ajlee21: So this is the similarity say between a single doc A from genetics vs doc B from chemistry? Where the documents are represented as the first 2 PCs?

ajlee21: AH...sorry i had been so used to you working with embeddings of your doc that i jumped to conclusions and missed the `,` before the PCs. I see now that you're comparing each doc to your PC axes...which leads me to Ben's same question of interpretation which you answered

biorxiv/pca_association_experiment/biorxiv_pca_category_bootstrap.py
ajlee21: This is much clearer now that I understand the earlier bits of code. Thanks

biorxiv/pca_association_experiment/biorxiv_pca_category_bootstrap.py
ajlee21: I see! I wonder if it would help to change the axis label to `similarity of document vs PC1` and maybe explain what PCs you're using in this case if PC1 = biochemistry in this case

PR #15, opened 2020-08-17
Updated corpora comparison

danich1:
This PR finalized the redo of the corpora comparison analysis I did a while back. I was told a way to improve this analysis is to incorporate a negative control corpus and to help highlight the differences I found in my last analysis. Turns out my negative control here is New York Times Annotated Corpus (NYTAC). Highlights for this PR are as follows:

1. Calculated general token/document statistics for bioRxiv, PMC and NYTAC 
2. Calculated odds ratios for top 200 words between bioRxiv. PMC, and NYTAC
3. Calculated the KL divergence between bioRxiv, PMC and NYTAC 

Biggest thing to review are the notebooks and figures with the figure parent directory,


annorxiver_modules/corpora_comparison_helper.py
```diff
@@ -0,0 +1,411 @@
...
+                    )
+                )
+            )
+        )
+    )
+    return ci_df
+
+def create_lemma_count_df(data_df, corpus_one_label, corpus_two_label):
+    """
+    Creates the dataframe that contains lemmas a the counts
```
ben-heil: What does this mean?

biorxiv/corpora_comparison/figure_generator.py
```diff
@@ -0,0 +1,409 @@
...
+
+
+# In[4]:
+
+
+g = (
+    p9.ggplot(
+        kl_divergence_df.replace({
+        "biorxiv_vs_nytac":"biorxiv_vs_reference",
+        "pmc_nytac": "pmc_vs_reference"
```
ben-heil: `pmc_vs_nytac` is still showing up on this figure which is confusing

```suggestion
        "pmc_vs_nytac": "pmc_vs_reference"
```

biorxiv/corpora_comparison/figure_generator.py
```diff
@@ -0,0 +1,409 @@
...
+)
+plot_df.head()
+
+
+# In[8]:
+
+
+g = plot_pointplot(plot_df, "bioRxiv vs PMC Odds Ratio")
+g.save("output/svg_files/biorxiv_pmc_frequency_odds.svg", dpi=75)
+print(g)
```
ben-heil, line 115: I saw "ml" in this figure and got excited, but it's probably millilitres here

biorxiv/corpora_comparison/figure_generator.py
```diff
@@ -0,0 +1,409 @@
...
+)
+plot_df.head()
+
+
+# In[15]:
+
+
+g = plot_pointplot(plot_df, "bioRxiv vs Reference Odds Ratio", True)
+g.save("output/svg_files/biorxiv_nytac_frequency_odds.svg", dpi=75)
+print(g)
```
ben-heil, line 199: Do you know what the 'u' that is the top ranked item is? Is that just a diacritical mark that goes on names?

ben-heil (approved):
Looks good to me! I left some minor comments and will defer to Alex on the KL-divergence/other stats implementation

biorxiv/corpora_comparison/word_frequency_counts.ipynb
```diff
@@ -12,39 +12,35 @@
...
   },
   {
    "cell_type": "code",
    "execution_count": 2,
    "metadata": {
     "ExecuteTime": {
-     "end_time": "2020-05-16T15:35:09.118438Z",
-     "start_time": "2020-05-16T15:35:08.605318Z"
+     "end_time": "2020-08-07T15:31:55.658944Z",
+     "start_time": "2020-08-07T15:31:55.122472Z"
```
ajlee21, line 44: Very minor, but just wanted to check
1. Do you need the extra `doi` column since that information is in the index? Perhaps you're using it some other place though
2. What does `disable = ['ner']` do?
3. Would it make sense to have cell 5 be made into a function since you're using it twice (for bioRiv and PMC)? Also your cells is labeled as `word frequency` which confused me a little because you're returning `sentence lengths`. But perhaps I'm misunderstanding something here. My assumption is that word frequency refers to the number of times the word `dog` appears versus sentence length is the number of words in a given sentence. Is that right?

annorxiver_modules/corpora_comparison_helper.py
```diff
@@ -0,0 +1,411 @@
...
+import numpy as np
+import pandas as pd
+import plotnine as p9
+import svgutils.transform as sg
+from tqdm import tqdm_notebook
+
+
+def aggregate_word_counts(doc_iterator):
+    """
+    This function aggregates the word count tsv files.
```
ajlee21, line 16: Might be helpful to have a mockup of what the word count tsv file looks like?
I imagine something like
```
| word | count |
|-------|--------|
"dog"  |  8       |
```

annorxiver_modules/corpora_comparison_helper.py
```diff
@@ -0,0 +1,411 @@
...
+        doc_iterator - a pathlib generator that returns file paths to be parsed
+    """
+    global_word_counter = Counter()
+    
+    for doc in tqdm_notebook(doc_iterator):
+        with open(doc, "r") as tsvfile:
+            reader = csv.DictReader(tsvfile, delimiter="\t")
+            global_word_counter.update({
+                (
+                    row['lemma'],
```
ajlee21, line 40: I'm not too sure what these tags are, do you think its worth adding a comment about that?

annorxiver_modules/corpora_comparison_helper.py
```diff
@@ -0,0 +1,411 @@
...
+    """
+    This function outputs the counter data object in
+    the form of a tsv file. This was created to avoid
+    the memory issue PMC creates.
+    
+    Arguments:
+        count_dict - the counter object to be outputted
+        file_name - the name of the file for the output to be written to
+    """
+    # Avoids memory issues
```
ajlee21: Not sure if you want to elaborate a bit more on the memory issue you had here. I can see this being potentially useful for future users of your awesome code!

annorxiver_modules/corpora_comparison_helper.py
```diff
@@ -0,0 +1,411 @@
...
+    with open(file_name, "w") as tsvfile:
```
ajlee21, line 129: Neat! I've been reading things in as pandas df and converting it to dict

biorxiv/corpora_comparison/corpora_comparison.py
```diff
@@ -0,0 +1,668 @@
...
+)
+nytac_sentence_length = pickle.load(
+    open("../../nytac/corpora_stats/nytac_sentence_length.pkl", "rb")
+)
+
+
+# In[8]:
+
+
+spacy_nlp = spacy.load('en_core_web_sm')
```
ajlee21, line 119: As a non-nlp person, is spacy a library that just gives you your stop words. I assume this is a pretty standard list used by people in your field?

biorxiv/corpora_comparison/corpora_comparison.py
```diff
@@ -0,0 +1,668 @@
...
+# ## Get Corpora Comparison Stats
+
+# In[9]:
+
+
+biorxiv_sentence_len_list = list(biorxiv_sentence_length.items())
+biorxiv_data = {
+    "document_count": len(biorxiv_sentence_length),
+    "sentence_count": sum(map(lambda x: len(x[1]), biorxiv_sentence_len_list)),
+    "token_count": biorxiv_total_count_df['count'].sum(),
```
ajlee21, line 132: Is it worth printing out an example `biorxiv_total_count_df` since you're using it to get various word distributions here?

biorxiv/corpora_comparison/corpora_comparison.py
```diff
@@ -0,0 +1,668 @@
...
+                    map(
+                        lambda x: x[1], 
+                        biorxiv_sentence_len_list
+                    )
+                )
+            )
+        )
+    ),
+    
+    "negatives":(
```
ajlee21, line 163: I keep learning new pandas functionality from you!

biorxiv/corpora_comparison/corpora_comparison.py
```diff
@@ -0,0 +1,668 @@
...
+stop_word_list = list(spacy_nlp.Defaults.stop_words)
+
+
+# ## Get Corpora Comparison Stats
+
+# In[9]:
+
+
+biorxiv_sentence_len_list = list(biorxiv_sentence_length.items())
+biorxiv_data = {
```
ajlee21, line 129: Not something you should do for this PR and I'm not sure what your timeline looks like moving forward. But if you think it makes sense to refactor, you could pull out as a function.

biorxiv/corpora_comparison/corpora_comparison.py
```diff
@@ -0,0 +1,668 @@
...
+        ['count']
+        .sum()
+    )/nytac_total_count_df['count'].sum()
+}
+
+
+# In[12]:
+
+
+token_stats_df = (
```
ajlee21, line 407: Maybe add a comment on what each row here means?

annorxiver_modules/corpora_comparison_helper.py
```diff
@@ -0,0 +1,411 @@
...
+        LL = lambda a,b,c,d: 2*(
+            a*np.log(a) + b*np.log(b) + c*np.log(c) + d*np.log(d)
+            - (a+b)*np.log(a+b) - (a+c)*np.log(a+c) - (b+d)*np.log(b+d)
+            - (c+d)*np.log(c+d) + (a+b+c+d)*np.log(a+b+c+d)
+        )
+        log_likelihood = LL(a,b,c,d)
+
+
+        # Log Odds
+        log_ratio = float((a*d)/(b*c))
```
ajlee21, line 219: Sorry I remember talking to you about the odds ratio representing the likelihood of seeing a word in one corpora versus another. But I'm getting a little lost with the total in your contingency table since usually there isn't a total... So you have (count in corpora 1/count in corpora 2)/(total count in corpora 1/total count in corpora 2)...So its the odds of seeing a word in corpora 1 vs seeing it in the total..?

annorxiver_modules/corpora_comparison_helper.py
```diff
@@ -0,0 +1,411 @@
...
+
+        a, b, c, d = (
+            observed_contingency_table[0][0],
+            observed_contingency_table[0][1],
+            observed_contingency_table[1][0],
+            observed_contingency_table[1][1]
+        )
+
+        # Obtained from (Kilgarriff, 2001) - Comparing Corpora
+        LL = lambda a,b,c,d: 2*(
```
ajlee21, line 210: I'm not sure if I follow what this likelihood is telling you?

annorxiver_modules/corpora_comparison_helper.py
```diff
@@ -0,0 +1,411 @@
...
+                    .assign(
+                        count=pseudo_count,
+                        freq=pseudo_count/corpus_top_total
+                    ),
+                )
+            )
+    
+    assert top_term_freq.shape[0] == bottom_term_freq.shape[0]
+    
+    kl_div = (
```
ajlee21, line 368: Looks good to me, so this is comparing the distribution of word counts between two corpora?

biorxiv/corpora_comparison/figure_generator.py
ajlee21: Why did you choose "nytac" as your reference. I guess I'm not sure what a "negative control" means in this context.

ajlee21: I assume that its a good validation that pmc vs bioriv are more similar compare to the reference, which covers a completely different area so we'd expect the language to be different..?

biorxiv/corpora_comparison/figure_generator.py
```diff
@@ -0,0 +1,409 @@
...
+    )
+    .replace('rna', 'RNA')
+)
+plot_df.head()
+
+
+# In[8]:
+
+
+g = plot_pointplot(plot_df, "bioRxiv vs PMC Odds Ratio")
```
ajlee21, line 113: Does it make sense that neurons and genome are at a high odds compared to PMC?

biorxiv/corpora_comparison/figure_generator.py
```diff
@@ -0,0 +1,409 @@
...
+    )
+    .replace('rna', 'RNA')
+)
+plot_df.head()
+
+
+# In[15]:
+
+
+g = plot_pointplot(plot_df, "bioRxiv vs Reference Odds Ratio", True)
```
ajlee21, line 197: Looks like notation has a higher odds in bioriv and pmc vs reference...that i think makes sense to me since they are writing about different topics. Was that your logic?

ajlee21 (approved):
This was such an interesting PR to read through. Looks like you've accomplished a lot.

All the code looks good. I just had a lot of comments to help clarify my understanding and possibly point to areas where you might want to reformat things (not for this PR) if you think its warranted.

annorxiver_modules/corpora_comparison_helper.py
danich1: It means that my comment has typos in it.  The correct form should be: creates a dataframe that contains lemmas (base form of words) and counts.

biorxiv/corpora_comparison/figure_generator.py
danich1: If I'm following correctly, I think that is the degree symbol. If it was a diacritical mark, I'd presume the letter would follow under it as the parser would leave those characters alone.

biorxiv/corpora_comparison/word_frequency_counts.ipynb
danich1: >Do you need the extra doi column since that information is in the index? Perhaps you're using it some other place though

Good point. It is redundant information as agg provides the doi in the index column; however, I usually ignore the index when I'm writing my code. This results in the index being reset eventually. 

>What does disable = ['ner'] do?

So Spacy's parser provides Named Entity Recognition as part of its parser. Since I only care about the part of speech tags and dependency tags, I disabled that feature to speed up computation.

> Would it make sense to have cell 5 be made into a function since you're using it twice (for bioRiv and PMC)? 

Good point. I'll reformat to do that.

> Also your cells is labeled as word frequency which confused me a little because you're returning sentence lengths. But perhaps I'm misunderstanding something here. My assumption is that word frequency refers to the number of times the word dog appears versus sentence length is the number of words in a given sentence. Is that right?

So I'm combining this function to be return word frequency, part of speech tag freq, dependency tag freq, as well as obtaining the length of every sentence within each document. This sentence length is for my comparison table that is needed for my manuscript. Basically I'm trying to do all the things in one go.

biorxiv/corpora_comparison/corpora_comparison.py
danich1: > As a non-nlp person, is spacy a library that just gives you your stop words. I assume this is a pretty standard list used by people in your field?

Yup. Spacy is pretty spiffy when it comes to handling NLP parsing as it provides a lot of useful NLP tools and lists. The stopword list is standard with plus or minus a few words here and there as researchers tweak the list.

biorxiv/corpora_comparison/corpora_comparison.py
danich1: >Is it worth printing out an example biorxiv_total_count_df since you're using it to get various word distributions here?

A few cells down I actually print out the full table.

biorxiv/corpora_comparison/corpora_comparison.py
danich1: >Not something you should do for this PR and I'm not sure what your timeline looks like moving forward. But if you think it makes sense to refactor, you could pull out as a function.

Hmmm.. I see what you are saying. I think I'll put that on the refactor list rather than change it as I don't think the function would be needed anywhere else in my repository.

annorxiver_modules/corpora_comparison_helper.py
danich1: > But I'm getting a little lost with the total in your contingency table since usually there isn't a total... So you have (count in corpora 1/count in corpora 2)/(total count in corpora 1/total count in corpora 2)...So its the odds of seeing a word in corpora 1 vs seeing it in the total..?

Almost got it. The total in this case is the total number of tokens in both corpora. The odds ratio for this is calculated as follows: (token count in corpora 1/total number of tokens in corpora 1)/((token count in corpora 2/total number of tokens in corpora 2). The odds ratio depicts how likely a token is to appear in one corpus vs another. So values > 1 indicate the token is more likely to appear in corpus one rather than corpus two.

annorxiver_modules/corpora_comparison_helper.py
danich1: > I'm not sure if I follow what this likelihood is telling you? 

The likelihood tells you how well does your model fit the data. By itself it is meaningless. You need to have at least two likelihood values to have any meaning. The bigger the value the better the fit.

annorxiver_modules/corpora_comparison_helper.py
danich1: >Looks good to me, so this is comparing the distribution of word counts between two corpora?

Bingo. Basically KL is comparing how similar are two corpora. based on the amount of tokens being compared. Low values mean both corpora are very similar and vice versa on high values.

biorxiv/corpora_comparison/figure_generator.py
danich1: >Why did you choose "nytac" as your reference. I guess I'm not sure what a "negative control" means in this context.

It was suggested by my committee member to use a non-biological related corpus as this would highlight large differences between corpora on a token level as well as the corpus distribution level (KL divergence).

> I assume that its a good validation that pmc vs bioriv are more similar compare to the reference, which covers a completely different area so we'd expect the language to be different..?

Yup you got it. It also makes the argument that PMC and bioRxiv are similar corpora, which is good to confirm for future experiments.

biorxiv/corpora_comparison/figure_generator.py
danich1: > Does it make sense that neurons and genome are at a high odds compared to PMC?

Yes. As the greatest difference between bioRxiv and PMC are topic specific terms related to neuroscience and bioinformatics. PMC contains those topics, but has a wider range of documents and textual content.

biorxiv/corpora_comparison/figure_generator.py
danich1: >Looks like notation has a higher odds in bioriv and pmc vs reference...that i think makes sense to me since they are writing about different topics. Was that your logic?

Yup, you got it. I'd be a bit surprised otherwise.

biorxiv/corpora_comparison/corpora_comparison.py
ajlee21: Sorry this question was not super clear. I guess what I was more asking was to see what the df looked like before your malnipulations, as I was interested in trying to learn more about how these pandas functions are working. But that was more for me..these are alot of chained pandas functions for me to digest:) If you have a screenshot of it for me to look at that might be helpful, but don't think you need to permanently change your code to print these df  out

danich1: Ah I gotcha. Will redirect this one offline as this is more centered around the package rather than code within this PR.

PR #16, opened 2020-08-17
Updated Publication Rate Measurement

danich1:
This PR is a follow up on the research [Abdill et al 2018](https://elifesciences.org/articles/45133) did to measure the rate of publication for bioRxiv preprints. Basically they measured the proportion of preprints published at month intervals from start of bioRxiv to Oct /2018. I updated this analysis to include the 2020 snapshot bioRxiv released to the publish. Plus, I incorporated a number of missing links thanks to my document representation distance method. 

Main files to view are the figure files, notebooks and notebook scripts.


biorxiv/article_distances/biorxiv_corrected_published_rate.py
```diff
@@ -0,0 +1,520 @@
...
+        if len(date_node) == 0:
+            print("Error in xml:")
+            print(document)
+            continue
+
+        elif len(date_node) < 2:
+            date_node = date_node[0]
+
+        else:
+            date_node = date_node[1]
```
ben-heil, line 191: What do these cases correspond to? Does biorxiv have two different date formats in the metadata?

biorxiv/article_distances/biorxiv_corrected_published_rate.py
```diff
@@ -0,0 +1,520 @@
...
+)
+print(f"Published: {published}")
+print(f"Posted: {posted}")
+print(f"Overall proportion published: {published/posted:.4f}")
+
+
+# In[20]:
+
+
+if False:
```
ben-heil: Is this a faster way of commenting out code? If so, do you still need these three cells?

biorxiv/article_distances/biorxiv_corrected_published_rate.py
```diff
@@ -0,0 +1,520 @@
...
+    )
+    
+    + p9.annotate(
+        "text", x=8.5, y=0.48, 
+        label=f"overall: {published/posted:.4f}", size = 8
+    )
+)
+g.save("output/figures/publication_rate.svg", dpi=500)
+g.save("output/figures/publication_rate.png", dpi=500)
+print(g)
```
ben-heil, line 456: This figure is super cool! Based on the way that both curves tail off it looks like after two years papers are pretty much published if they ever will be.

Also I feel like the overall proportion published lines are kind of hard to interpret. I read them as "because there have been more preprints over time the proportion published in 2020 is higher than in 2018 since ~75 percent get published regardless." I know that's not necessarily true though because each point on the chart corresponds to a different number of preprints/publications, which could skew the lines up or down as well.

ben-heil (approved):
Looks good! I left a couple comments on the main figure

biorxiv/article_distances/biorxiv_corrected_published_rate.py
```diff
@@ -0,0 +1,520 @@
...
+
+
+file_path="https://elifesciences.org/download/aHR0cHM6Ly9jZG4uZWxpZmVzY2llbmNlcy5vcmcvYXJ0aWNsZXMvNDUxMzMvZWxpZmUtNDUxMzMtZmlnMy1kYXRhMS12Mi5jc3Y-/elife-45133-fig3-data1-v2.csv?_hash=6d5MoXPaQXFZiifDTRaKXoZ29OnVUJq%2FuoHlyuh%2Bg04%3D"
+
+
+# In[3]:
+
+
+abdill_df = pd.read_csv(file_path)
+abdill_df.head(10)
```
ajlee21, line 30: Just curious, what do these integer values represent in this table? I assume its a temporal representation but not clear how it was generated.

biorxiv/article_distances/biorxiv_corrected_published_rate.py
```diff
@@ -0,0 +1,520 @@
...
+
+
+snapshot_wo_links_df = (
+    biorxiv_journal_df
+    .assign(published_closed=lambda x: (x.pmcoa != x.published))
+    .merge(
+        final_mapped_df[["document", "pub_month"]],
+        on="document"
+    )
+    .replace({"2013-10":"2013-11"})
```
ajlee21, line 268: Why are you replacing this here? Seems like a one-off thing.

biorxiv/article_distances/biorxiv_corrected_published_rate.py
```diff
@@ -0,0 +1,520 @@
...
+        linetype="solid", 
+        color=color_mapper['2020ML']
+    )
+    
+    + p9.annotate(
+        "text", x=8.5, y=0.395, 
+        label="overall: 0.4196", size = 8
+    )
+    
+    + p9.annotate(
```
ajlee21, line 449: Nice use of plotnine functions!

biorxiv/article_distances/biorxiv_corrected_published_rate.py
ajlee21: Agree that this plot is very cool to see! I am also a little confused by  some of the trends...so my take on this plot is that you have proportion of preprints that get published on your y-axis and time on the x. When the curve trailed off, I initially thought that that mean that many recent preprint were not getting published. However, I realized that this trailing off is likely because these preprints are still in review. I'm not sure how to best account...not sure if "in review" is available as metadata anywhere.

I'm not following Ben's point but perhaps your comment will clarify things

Last, when you calculate missing links. Are these preprints that you imputed were published from similarity distance to a published paper, or are these the preprints where you manually found the actual published paper where the link was missing?

ajlee21 (approved):
Looks good to me, just a couple of comments

biorxiv/article_distances/biorxiv_corrected_published_rate.py
mthielk: Nice fig!
I'm trying to mesh your results with the figure I made:
image
This is theoretically a non parametric method and I've plotted error bars (they're just tiny for all of biorxiv). X axis is time since first shared on biorxiv.
I'm not sure why my overall publication rate settles to such a different value than your figure. This is using data that is pretty equivalent to your crossref mapping I believe. I haven't had a chance to look at your new embedding based method, but your mapping makes it look like it's not that big of a difference?
I haven't been particularly careful about versions or retractions. Do you do anything regarding these that might account for the differences?

mthielk: oh, for figure aesthetics/to address some of the other comments. Perhaps vertical lines to represent error bars assuming a bionomial process might give a sense of publication volume over time and which deviations are significant... But this could also dilute the main point of the figure so ¯\_(ツ)_/¯

biorxiv/article_distances/biorxiv_corrected_published_rate.py
danich1: > What do these cases correspond to? Does biorxiv have two different date formats in the metadata?

BioRixv has a submitted date and accepted date. I'm using the date the preprints were submitted for my analysis. As with all things though there are errors within the bioRxiv xml files and I needed to account for that.

biorxiv/article_distances/biorxiv_corrected_published_rate.py
danich1: > Is this a faster way of commenting out code? If so, do you still need these three cells? 

This code here was being used to try and estimate the true number of missing biorxiv-pmc links; however, it was determined that the equation isn't needed anymore. I meant to delete it but forgot.

biorxiv/article_distances/biorxiv_corrected_published_rate.py
danich1: > so my take on this plot is that you have proportion of preprints that get published on your y-axis and time on the x.

You are correct.

> When the curve trailed off, I initially thought that that mean that many recent preprint were not getting published. However, I realized that this trailing off is likely because these preprints are still in review. I'm not sure how to best account...not sure if "in review" is available as metadata anywhere.

There's no way to account for that unless the authors post an open review. I do not know the stats for the number of open reviews; however, I don't think it is that many. 

>"because there have been more preprints over time the proportion published in 2020 is higher than in 2018 since ~75 percent get published regardless." I know that's not necessarily true though because each point on the chart corresponds to a different number of preprints/publications, which could skew the lines up or down as well.

>I'm not following Ben's point but perhaps your comment will clarify things

Basically, Ben is saying that the overall number of preprints being published is higher in 2020 than in 2018. This is true as enough time has past for preprints to be published; however, the catch here is we don't know the true proportion  of preprints being published as papers are published behind paywalls etc. So trying to compare the overall proportion published (horizontal lines), is not easy as there are too many co-founders to 100% agree with the gap.

>Last, when you calculate missing links. Are these preprints that you imputed were published from similarity distance to a published paper, or are these the preprints where you manually found the actual published paper where the link was missing?

The links were imputed via my similarity distance metric and then classified after I manually checked to see if bioRxiv already contained the published link.

danich1: > I'm not sure why my overall publication rate settles to such a different value than your figure. This is using data that is pretty equivalent to your crossref mapping I believe. I haven't had a chance to look at your new embedding based method, but your mapping makes it look like it's not that big of a difference?I haven't been particularly careful about versions or retractions. Do you do anything regarding these that might account for the differences?

The reason ours looks different is that my estimate includes my imputed missing links. This pushes the overall proportion up by about 5%. If you take a look at the 2020 Snapshot curve it should be similar to what you are plotting.

>Perhaps vertical lines to represent error bars assuming a bionomial process might give a sense of publication volume over time and which deviations are significant... But this could also dilute the main point of the figure so ¯_(ツ)_/¯

Thanks for the suggestion. Will make a new PR if this pans out.

biorxiv/article_distances/biorxiv_corrected_published_rate.py
danich1: > Just curious, what do these integer values represent in this table? I assume its a temporal representation but not clear how it was generated.

The integers represent the number of preprints posted (posted) and the number of preprints published (published). The authors binned preprints by month and then asked how many of these preprints are published via biorxiv's api.

biorxiv/article_distances/biorxiv_corrected_published_rate.py
danich1: > Why are you replacing this here? Seems like a one-off thing.

I replaced these values to be in line with Abdill et al's results. They start at 2013-11 instead of 2013-10.

PR #17, opened 2020-09-08
Corpora comp quick fix

danich1:
This PR is a quick fix for calculating the odds ratio to compare preprints against their published versions. 
On the master branch preprint_published_comparison.png is the same as the global comparison figure. Reason for this is I forgot to update my code to reflect the new changes and I have now fixed that.
No real review necessary as the only main change is the figure files.


biorxiv/corpora_comparison/figure_generator.py
```diff
@@ -356,7 +356,7 @@
     .append(
         full_plot_df
         .sort_values("odds_ratio", ascending=False)
-        .iloc[:-2]
+        .iloc[:-3]
```
ben-heil, line 359: Is the `iloc` call here to manually get rid of infinite or NaN odds ratios? If so, would it make more sense to do it explicitly so you aren't manually hiding rows?

ben-heil (approved):
Looks good, I only left a minor comment for consideration

biorxiv/corpora_comparison/corpora_comparison.ipynb
```diff
@@ -2779,8 +2799,8 @@
    ],
    "source": [
     "preprint_vs_published = get_term_statistics(\n",
-    "    biorxiv_total_count_df, \n",
-    "    pmc_total_count_df, \n",
+    "    preprint_count_df, \n",
```
ajlee21, line 2802: I assume this is where you over-wrote your figure so they looked the same?

ajlee21 (approved):
LGTM

biorxiv/corpora_comparison/figure_generator.py
danich1: >Is the iloc call here to manually get rid of infinite or NaN odds ratios? If so, would it make more sense to do it explicitly so you aren't manually hiding rows?

Actually no. The reason I have this line here is to remove non-informative tokens. For example the true leading token in this dataframe is 'u\2009'. This is a unicode for thin space character.  Not informative at all, so it needed to be removed.

biorxiv/corpora_comparison/corpora_comparison.ipynb
danich1: Correct I accidentally passed the wrong dataframes for this section of code. Hence the duplication.

biorxiv/corpora_comparison/figure_generator.py
ben-heil: Ahhh I see

PR #19, opened 2020-09-17
Upload Journal Recommendation Figures

danich1:
This PR contains code for generating figures to be used by the journal recommender web resource. Plus, this PR contains a figure that I'll be using for the manuscript that I am currently writing. Main code review looking at the notebook scripts as the generated figures are designed for debugging purposes only.


pmc/journal_recommendation/journal_recommendation_figure_generator.py
```diff
@@ -0,0 +1,570 @@
...
+
+
+# In[2]:
+
+
+# Set up porting from python to R
+# and R to python :mindblown:
+
+import rpy2.rinterface
+get_ipython().run_line_magic('load_ext', 'rpy2.ipython')
```
ben-heil, line 35: <3 rpy2

pmc/journal_recommendation/journal_recommendation_figure_generator.py
```diff
@@ -0,0 +1,570 @@
...
+    {"value":.39982, "model":"paper_paper", "distance":"euclidean", "dataset":"train"},
+    {"value":.37340, "model":"centroid", "distance":"euclidean", "dataset":"train"},
+    {"value":.39824, "model":"paper_paper", "distance":"manhattan", "dataset":"train"},
+    {"value":.39824, "model":"centroid", "distance":"manhattan", "dataset":"train"},
+    {"value":.00445, "model":"random_baseline", "distance":"N/A", "dataset":"test"},
+    {"value":.20638, "model":"paper_paper", "distance":"euclidean", "dataset":"test"},
+    {"value":.21636, "model":"centroid", "distance":"euclidean", "dataset":"test"},
+    {"value":.20380, "model":"paper_paper", "distance":"manhattan", "dataset":"test"},
+    {"value":.21523, "model":"centroid", "distance":"manhattan", "dataset":"test"}
+]
```
ben-heil, line 54: Where do these results come from? Should they be read from a file instead of hardcoded?

pmc/journal_recommendation/square_bins_word_ratios.py
```diff
@@ -0,0 +1,203 @@
...
+    
+    agg_word_count = aggregate_word_counts(files, disable_progressbar=True)
+    
+    filtered_agg_word_count = {
+       term[0]:agg_word_count[term] 
+        for term in agg_word_count 
+        if term[1] != 'SPACE' and term[0] not in stop_word_list
+    }
+    
+    global_word_counter.update(Counter(filtered_agg_word_count))
```
ben-heil: Is there a reason to cast the `dict` to a `Counter` before passing it into update?

ben-heil (approved):
Looks good!

pmc/journal_recommendation/journal_recommendation_figure_generator.py
```diff
@@ -0,0 +1,570 @@
...
+)
+
+result_df.head()
+
+
+# In[5]:
+
+
+g = (
+    p9.ggplot(result_df.query("distance in ['euclidean', 'N/A']"), p9.aes(x="model", y="value"))
```
ajlee21, line 74: So there's not a difference between recommending using the dist(your paper, paper in journal X) vs dist(your paper, centroid of all papers in journal X)?

ajlee21: Does that mean that papers within a journal are pretty robust/similar? I thought you mentioned these big journals with a lot of different topics..maybe these were removed?

pmc/journal_recommendation/journal_recommendation_figure_generator.py
```diff
@@ -0,0 +1,570 @@
...
+    full_paper_dataset_subset,
+    lambda_c_grid=lambda_c_grid,
+    lambda_d_grid=lambda_d_grid
+)
+
+
+# In[13]:
+
+
+g = (
```
ajlee21, line 238: Maybe add a quick description of these plots. So these are plots of papers from your 4 different journals using different params of saucie? Whats the takeaway for which params to use -- I assume the one that clusters by journal? Are these PC dims?

pmc/journal_recommendation/journal_recommendation_figure_generator.py
```diff
@@ -0,0 +1,570 @@
...
+    sep="\t", index=False
+)
+
+full_dataset.head()
+
+
+# In[22]:
+
+
+g = (
```
ajlee21, line 389: Since this plot is going in your app, maybe rename the axes to be more interpretable..?

ajlee21 (approved):
Code all looks good to me! Just a couple of questions

pmc/journal_recommendation/journal_recommendation_figure_generator.py
danich1: These results come from the journal_recommendation_experiment notebook. It was easier to hard code the results than rerun that notebook as it can take up to days to run. This situation it makes sense to just let it be, but good to note for the future.

pmc/journal_recommendation/square_bins_word_ratios.py
danich1: No hard reason for it being casted as a Counter. Just helps orient me programming wise to know that I'm working with a counter rather than pure dictionary.

pmc/journal_recommendation/journal_recommendation_figure_generator.py
danich1: >So there's not a difference between recommending using the dist(your paper, paper in journal X) vs dist(your paper, centroid of all papers in journal X)?

Test set wise there isn't a strong difference between both classifiers.

>Does that mean that papers within a journal are pretty robust/similar? I thought you mentioned these big journals with a lot of different topics..maybe these were removed?

No I don't think this result supports that claim. The reason for decent performance is that the mega-journals are being classified correctly for most papers that originated from them. If you take the mega-journals out performance drops.

pmc/journal_recommendation/journal_recommendation_figure_generator.py
danich1: >Since this plot is going in your app, maybe rename the axes to be more interpretable..?

For the notebook this wont be necessary as the journal recommendation app doesn't require axis labels. The dimension names here are pretty arbitrary as the axes don't really have an interpretable meaning. Good point for future experiments to keep in mind.

PR #21, opened 2020-11-03
Publication experiment

danich1:
This PR contains experiments on estimating the publication delay/determining which category has the fastest publication rate. I also updated the preprint to published mapper so I can have more accurate results. The time to publication is a reburish of Marvin's lifeline curves.


mthielk:
Cool analysis! I didn't think about looking at number of versions. I guess you could also use different preprint versions as a gold set for your distance/embedding matching algorithm (I'm not sure if you already did that or already found a good way to set the cutoff for the matches).

If you're interested in including the version_count_vs_publication_time figure in the manuscript, you might want to consider something like a [ridgeplot](https://seaborn.pydata.org/examples/kde_ridgeplot.html) where you do a KDE plot for each version count. Mostly because I've always wanted an excuse to do a ridgeplot but never had a good use. I could help make the figure if you want.

I might have also chosen a [hexbin plot](https://seaborn.pydata.org/examples/hexbin_marginals.html) for the article_distance_vs_publication_time figure, mainly cause I like hexagons and I'm more of a seaborn guy than a plotnine guy.

One thing I thought might be interesting is seeing if any of your principal component dimensions are strongly correlated with time to publication. Or even coloring your embedded documents with time to publication.

For the time_to_published_refurbished nb and half life analysis, it might also be worthwhile to calculate the half life for papers that eventually get published which would get around some categories that never reach 50% publication rate.

Let me know if you want help with anything.
Thanks,
~M

biorxiv/journal_tracker/map_preprints_to_journals.py
```diff
@@ -95,112 +83,89 @@ def call_pmc(doi_ids):
...
+    response = call_biorxiv(dois[batch:batch+batch_limit])
     doi_mapper_records += [
         {
-         "preprint_doi":result['message']['DOI'],
-         "published_doi":(
-                 result['message']['relation']['is-preprint-of'][0]['id'] 
-                 if "is-preprint-of" in result['message']['relation'] 
-                 else ""
-             )
+            "preprint_doi": collection['doi'],
```
ajlee21, line 99: Just to check my understanding. So you're 1) mapping preprint doi >> published dois, 2) using published dois to get journal and date published infor, 3) then you get PMC id to fill in missing links (should this step be done earlier? Or do you need the journal information to make this prediction?)

biorxiv/time_to_publication/time_to_published_refurbished.py
```diff
@@ -0,0 +1,240 @@
...
+    .assign(label="all_papers")
+)
+overall_preprint_survival.head()
+
+
+# In[9]:
+
+
+g = (
+    p9.ggplot(overall_preprint_survival, p9.aes(x="timeline", y="KM_estimate", color="label"))
```
ajlee21: Maybe adding `(days to publication)` for the axis label. I was a bit confused on how to interpret `timeline`

biorxiv/time_to_publication/time_to_published_refurbished.py
```diff
@@ -0,0 +1,240 @@
...
+            .reset_index()
+            .assign(category=cat)
+        )
+    )
+
+
+# In[11]:
+
+
+g = (
```
ajlee21, line 186: Nice plot! So how would you explain the shape of the curve? Is the idea that for all categories, there are papers that quickly move from preprint to publication, but there are a portion that take a while? Or are these preprints those that are in progress that haven't been submitted to a journal? If that is the case, is there a way to correct for these preprints, as I feel like its skewing the results if the takeaway is to get a measure for how quickly preprints move to publication.

biorxiv/publication_delay_experiment/publication_delay_experiment.py
```diff
@@ -0,0 +1,516 @@
...
+)
+
+print(
+    f"Biorxiv Stats Mean: {np.mean(biorxiv_distances):.3f}, "
+    f"Std: {np.std(biorxiv_distances):.3f}"
+)
+
+
+# Take home results:
+#     1. Preprints are delayed 51 days for each new version that needs to be created
```
ajlee21: Maybe just a wording issue, but this statement makes me think that publishing more will delay time to publication, when my take is that its not the number of versions that is necessarily causing this increase in time to publication but there is a correlation between the two where if you're having trouble publishing you're releasing more versions..

biorxiv/publication_delay_experiment/publication_delay_experiment.py
```diff
@@ -0,0 +1,516 @@
...
+    .set_index("document")
+    .loc[matched_preprint_published_pairs.document.tolist()]
+)
+biorxiv_documents.head()
+
+
+# In[11]:
+
+
+pmc_documents = (
```
ajlee21, line 198: Just to clarify these are full text embeddings, when you're calculating the distance?

ajlee21 (approved):
Really cool plots!

Just had a few questions and comments.

On a minor note, if you're planning to use these figures for your manuscript, might want to update the axis labels to remove the "_", maybe also increase the size of the figures too.

biorxiv/journal_tracker/map_preprints_to_journals.py
```diff
@@ -95,112 +83,89 @@ def call_pmc(doi_ids):
...
-    .drop("preprint_doi", axis=1)
+    .assign(
+        version=lambda x: x.document.apply(lambda doc: int(doc.split(".")[0][-1])),
+    )
+    .rename(index=str, columns={"doi":"preprint_doi"})
+    .merge(
+        published_doi_df.assign(
+            published_doi=lambda x: x.published_doi.apply(
+                lambda url: re.sub(r"http(s)?://doi.org/", '', url) 
+                if type(url) == str else url
```
ben-heil, line 145: What type would `url` be other than string? Is this condition catching regexes that didn't have an https header or preventing attempts of regexes on non-string urls?

biorxiv/journal_tracker/map_preprints_to_journals.py
```diff
@@ -240,39 +206,45 @@ def call_pmc(doi_ids):
...
-# In[12]:
+# In[10]:
 
 
 chunksize=100
 data = []
 for chunk in tqdm.tqdm(range(0, len(missing_ids), chunksize)):
     query_ids = missing_ids[chunk:chunk+chunksize]
-    response = call_pmc(query_ids).json()
+    response = call_pmc(query_ids, 'annorxivr', 'dnicholson329@gmail.com').json()
```
ben-heil: Do you want to pull your email from a config file, or is it okay to have it here since it's in a notebook?

biorxiv/publication_delay_experiment/publication_delay_experiment.py
```diff
@@ -0,0 +1,516 @@
...
+
+# In[15]:
+
+
+g = (
+    p9.ggplot(
+        published_date_distances, 
+        p9.aes(y="time_to_published", x="version_count")
+    )
+    + p9.geom_point()
```
ben-heil: It might be easier to make sense of the publication date distribution if you use geom_violin or geom_boxplot here

ben-heil: Oh, just saw marvin's comment about a ridge plot, that would work too

biorxiv/publication_delay_experiment/publication_delay_experiment.py
```diff
@@ -0,0 +1,516 @@
...
+    )
+    + p9.geom_bin2d(bins=100)
+    + p9.geom_line(
+        stat="smooth",method='lm', linetype='dashed', 
+        se=False, alpha=0.9, size=0.6
+    )
+    + p9.scale_y_timedelta(labels=timedelta_format('d'))
+    + p9.annotate(
+        'text', x=10, y=timedelta(days=1450), 
+        label=f"Y={results_2.slope:.2f}*X+{results_2.intercept:.2f}"
```
ben-heil: If you decide to map the densities to color, you may want to use the log of the counts instead of the raw counts to prevent the large values from washing everything out

biorxiv/time_to_publication/time_to_published_refurbished.py
```diff
@@ -0,0 +1,240 @@
...
+            .tolist()
+        )
+    )
+    + p9.coord_flip()
+    + p9.labs(
+        x="preprint category",
+        y="Days Until 50% of Preprints are Published",
+        title="Preprint Category Half-Life"
+    )
+)
```
ben-heil, line 263: I love the last two plots here! It may be possible to combine the information by annotating the outliers on the half life curves though, since everything else appears to be roughly the same.

biorxiv/time_to_publication/time_to_published_refurbished.py
```diff
@@ -0,0 +1,240 @@
...
+g.save("output/preprint_category_halflife.svg", dpi=500)
+g.save("output/preprint_category_halflife.png", dpi=500)
+print(g)
+
+
+# Take home Results:
+#     1. The average amount of time for half of all preprints to be published is 348 days (~1 year)
+#     2. Biophysics and biochemistry are two categories that take the least time to have half their preprints published
```
ben-heil, line 271: I could see a short half life meaning either 'papers in this field don't take long to get published' or 'authors in this field submit a preprint once their manuscript is already accepted/close to it'. Is there a relationship between document distance and time to publish in particular fields?

ben-heil (approved):
Very cool!

biorxiv/journal_tracker/map_preprints_to_journals.py
danich1: `url` here could be None as some preprints do not have published counterparts. This section is used to remove malformed doi entries so I can use it for down stream analyses later. e.g. Should be 10.1101/#### but I get http://doi.org/10.1101/#### or even http://nature.org/####/##### <- not a doi.

biorxiv/journal_tracker/map_preprints_to_journals.py
danich1: > Do you want to pull your email from a config file, or is it okay to have it here since it's in a notebook?

Ack thanks for catching. Yeah I need to pull that out. Don't want random users frauding me lol.

biorxiv/publication_delay_experiment/publication_delay_experiment.py
danich1: > Just to clarify these are full text embeddings, when you're calculating the distance?

Correct. I'm using full text embeddings to calculate the distance between a preprints first version and a preprints published version.

biorxiv/journal_tracker/map_preprints_to_journals.py
danich1: >Just to check my understanding. So you're 1) mapping preprint doi >> published dois, 2) using published dois to get journal and date published infor, 3) then you get PMC id to fill in missing links (should this step be done earlier? Or do you need the journal information to make this prediction?)

So I'm mapping preprint dois to published dois, then I'm mapping published dois to PMCIDs. Lastly, I'm collecting the journal info thats connected to the corresponding PMCIDs. Need to do it in this order so I can make sure I have the correct PMCIDs.

biorxiv/time_to_publication/time_to_published_refurbished.py
danich1: >Maybe adding (days to publication) for the axis label. I was a bit confused on how to interpret timeline

Good point. I'll be converting these figures into publication ready versions so I'll keep that in mind :).

biorxiv/time_to_publication/time_to_published_refurbished.py
danich1: >So how would you explain the shape of the curve? Is the idea that for all categories, there are papers that quickly move from preprint to publication, but there are a portion that take a while?  Or are these preprints those that are in progress that haven't been submitted to a journal? If that is the case, is there a way to correct for these preprints, as I feel like its skewing the results if the takeaway is to get a measure for how quickly preprints move to publication.

You got it. The steeper the curve the more quickly preprints in that category are being published.

You are also correct as there are preprints that have yet to be published. It's okay if those unpublished preprints are skewing these curves as it is common knowledge that not all preprints get published. It portrays a realistic view on preprint-publication success.

biorxiv/publication_delay_experiment/publication_delay_experiment.py
danich1: >Maybe just a wording issue, but this statement makes me think that publishing more will delay time to publication, when my take is that its not the number of versions that is necessarily causing this increase in time to publication but there is a correlation between the two where if you're having trouble publishing you're releasing more versions..

Valid point. I'll reword this statement. Had trouble with phrasing this point and I think you just handed me a golden egg. Thanks!!

biorxiv/time_to_publication/time_to_published_refurbished.py
danich1: >Is there a relationship between document distance and time to publish in particular fields? 

Great question. This would be a great follow up to my overall doc distance to time published figure (publication_delay_experiment notebook). Will look into it, but hypothesis here is that some categories would have a lower slope than other categories if I were to perform linear regression for each individual category.

PR #27, opened 2021-01-06
Installed pre-commit checker

danich1:
This is a small pre-commit hook I added to the repository. This will make future commits pre-formatted by black and make it easier for me to remain consistent on how I code.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,21 @@
+repos:
+
+- repo: local
+  hooks:
+  - id: jupytext_format
+    name: jupytext_format
+    entry: jupytext --from ipynb --pipe black --check flake8 --pre-commit
+    pass_filenames: false
+    language: python
+
+  - id: jupytext_conversion
+    name: jupytext_conversion
+    entry: jupytext --from ipynb --to py:light --pre-commit
+    pass_filenames: false
+    language: python
+
+- repo: https://gitlab.com/pycqa/flake8
+  rev: 3.7.9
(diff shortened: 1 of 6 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,21 @@
+repos:
+
+- repo: local
+  hooks:
+  - id: jupytext_format
+    name: jupytext_format
+    entry: jupytext --from ipynb --pipe black --check flake8 --pre-commit
+    pass_filenames: false
+    language: python
```
ben-heil, line 9: Oooh I didn't know that jupytext could also run linters on notebooks

ben-heil (approved):
This is cool, I need to do something like this for formatting my notebooks

.pre-commit-config.yaml
danich1: Yeah I found this gem, which makes it so much easier to remain consistent with my code.

cgreene: This is amazing. This should get spammed all over our slack!

ajlee21: cool!

install.sh
```diff
@@ -0,0 +1,5 @@
+conda env create --file environment.yml
```
ajlee21, line 1: Really nice idea to keep the installation simple for users. I should add this to my repos as well!

ajlee21 (approved):
Really nice additions to make the repo a lot cleaner! Thanks for sharing the great tips!

PR #28, opened 2021-01-22
Update figures

danich1:
This PR doesn't really need review just updated my pre-commit checker to work with an updated version and updated figure per thesis committee feedback.


ben-heil (approved):
I felt a great disturbance in the force. As if hundreds of formatting irregularities cried out in terror and were suddenly silenced. Also the KL-divergence figure (I think that's the new one) looks good!

annorxiver_modules/corpora_comparison_helper.py
```diff
@@ -1,81 +1,92 @@
...
             # dep_tag - the dependency path tag
-            global_word_counter.update({
-                (
-                    row['lemma'], 
-                    row['pos_tag'], 
-                    row['dep_tag']
-                ):int(row['count'])
-                for row in reader
-            })
+            global_word_counter.update(
```
ajlee21, line 42: woah...cool new pandas function i havesn't seen!

annorxiver_modules/corpora_comparison_helper.py
```diff
@@ -115,86 +121,79 @@ def dump_to_dataframe(count_dict, file_name):
...
-        set(
-            corpus_two
-            .sort_values("count", ascending=False)
-            .head(freq_num)
-            .lemma
-            .values
-        )
+    term_list = set(
+        corpus_one.sort_values("count", ascending=False).head(freq_num).lemma.values
+    ) | set(
```
ajlee21, line 169: cool didn't know you could use this shorthand operator!

ajlee21 (approved):
Looks good. Like the new figure too, like the colors!

Just a quick question since its been a  little while since I looked at your PRs. Can you interpret your new plot to mean that as you increase the number of terms (i.e. size of your vocabulary) the difference increases between your training sets. So you're seeing more of a difference between the types of papers that your corpora contains as you increase the number of words you consider..?

annorxiver_modules/corpora_comparison_helper.py
danich1: That's actually a dictionary function. Not pandas fyi

annorxiver_modules/corpora_comparison_helper.py
danich1: Yeah its great to utilize!

danich1:
>Just a quick question since its been a little while since I looked at your PRs. Can you interpret your new plot to mean that as you increase the number of terms (i.e. size of your vocabulary) the difference increases between your training sets. So you're seeing more of a difference between the types of papers that your corpora contains as you increase the number of words you consider..?

Yes as the number of terms increases the difference between corpora word frequency increases. This makes sense as more words are being considered. Overall it shows that bioRxiv and PMC are more similar than NYTAC which is what one expects.

PR #29, opened 2021-03-29
Analyzing Preprints in Motion

danich1:
This PR uploads results based on work done within [this preprint](https://www.biorxiv.org/content/10.1101/2021.02.20.432090v1). Long story short this work complements the work I'm doing, so it is the perfect opportunity to do a joint submission. As a result I needed to run their subset through my pipeline to have results/update my manuscript hence this PR.  I also added some figure 1 nitpicking changes into the PR. 

You can ignore pretty much every file except for the following:
01_download_polka_subset_experiment.ipynb (light review as it is mainly processing files)
02_get_term_counts_and_doc_vectors.ipynb (light review as it is mainly processing files)
03_rerun_analyses_with_added_subset.ipynb (Review definitely needed)
Files within biorxiv/polka_subset_experiment/output/figures/*png


biorxiv/polka_subset_experiment/01_download_polka_subset_experiment.py
```diff
@@ -0,0 +1,153 @@
...
+import tqdm
+import urllib.request as request
+# -
+
+manual_papers_df = pd.read_csv(str(Path("output/all_pairs_2021-02-11.csv")))
+manual_papers_df.head().T
+
+manual_papers_df.exclude.unique()
+
+papers_to_download = manual_papers_df.fillna("").query(
```
ajlee21, line 36: Ah this is cleaner than doing `manual_papers_df[manual_papers_df["exclude"].contains...]` Nicce!

biorxiv/polka_subset_experiment/01_download_polka_subset_experiment.py
```diff
@@ -0,0 +1,153 @@
...
+#       format_name: light
+#       format_version: '1.5'
+#       jupytext_version: 1.9.1+dev
+#   kernelspec:
+#     display_name: Python [conda env:annorxiver]
+#     language: python
+#     name: conda-env-annorxiver-py
+# ---
+
+# # Grab the bioRxiv subset from Polka et al 2021
```
ajlee21, line 16: I know you're in a rush with this analysis, so this doesn't need to be done in this PR. But it might be helpful to have a short description for what this notebook is doing (i.e. This notebook is downloading what data from Polka et. al. and processing it to get what data)

biorxiv/polka_subset_experiment/01_download_polka_subset_experiment.py
```diff
@@ -0,0 +1,153 @@
...
+
+# # Download Files from PMCOA's FTP server
+
+pmc_open_access_url = "ftp://ftp.ncbi.nlm.nih.gov/pub/pmc/oa_bulk/"
+response = request.urlopen(f"{pmc_open_access_url}")
+files = response.read().decode("utf-8").splitlines()
+tar_files = [f.split(" ")[-1] for f in files]
+pmcid_list = pmcid_mapped_papers_df.PMCID.tolist()
+
+if not any(Path("output/pmcoa_xml_files").iterdir()):
```
ajlee21, line 122: For this kind of querying, do you have checks in place in case it crashes in the middle?

biorxiv/polka_subset_experiment/02_get_term_counts_and_doc_vectors.py
```diff
@@ -0,0 +1,118 @@
...
+#     text_representation:
+#       extension: .py
+#       format_name: light
+#       format_version: '1.5'
+#       jupytext_version: 1.9.1+dev
+#   kernelspec:
+#     display_name: Python [conda env:annorxiver]
+#     language: python
+#     name: conda-env-annorxiver-py
+# ---
```
ajlee21, line 14: Similar note about adding a description here. So this notebook is getting how many times a word appeared in all biorxiv and all pmcoa. This notebook is also returning document embeddings/vectors for abstract documents in biorxiv and pmcoa.

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
```diff
@@ -0,0 +1,603 @@
...
+    .groupby("lemma")
+    .agg({"count": "sum"})
+    .reset_index()
+    .sort_values("count", ascending=False)
+)
+
+preprint_count_df.head()
+# -
+
+# ## PMCOA Token Counts
```
ajlee21, line 119: Token  = lemma?

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
```diff
@@ -0,0 +1,603 @@
...
+polka_published = list(Path("output/pmcoa_word_counts").rglob("*tsv"))
+
+# +
+published_count = aggregate_word_counts(polka_published)
+
+published_count_df = (
+    pd.DataFrame.from_records(
+        [
+            {
+                "lemma": token[0],
```
ajlee21, line 130: Looks like lemma is part of a token. I guess I thought token = word. I know I have asked this before, please remind me again

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
```diff
@@ -0,0 +1,603 @@
...
+    .reset_index()
+    .sort_values("count", ascending=False)
+)
+
+published_count_df.head()
+# -
+
+# ## Get Token Stats
+
+preprint_vs_published = get_term_statistics(preprint_count_df, published_count_df, 100)
```
ajlee21: Can you describe the columns here. I assume that the corpus columns are the counts for that lemma in each corpus, but I'm not sure which ones. And then the LL is the likelihood of that lemma for something?

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
```diff
@@ -0,0 +1,603 @@
...
+        odds_ratio=lambda x: x.odds_ratio.apply(lambda x: np.log2(x)),
+        lower_odds=lambda x: x.lower_odds.apply(lambda x: np.log2(x)),
+        upper_odds=lambda x: x.upper_odds.apply(lambda x: np.log2(x)),
+    )
+)
+plot_df.head()
+
+g = (
+    p9.ggplot(
+        plot_df, p9.aes(y="lemma", x="lower_odds", xend="upper_odds", yend="lemma")
```
ajlee21, line 174: What are the upper and lower odds?

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
```diff
@@ -0,0 +1,603 @@
...
+    )
+    + p9.annotate("text", label="Preprint Enriched", x=-5, y=38.5, size=12, alpha=0.7)
+    + p9.theme_seaborn(context="paper", style="ticks", font_scale=1.1, font="Arial")
+    + p9.theme(
+        figure_size=(10, 6),
+        panel_grid_minor=p9.element_blank(),
+    )
+    + p9.labs(y=None, x="Preprint vs Published log2(Odds Ratio)")
+)
+g.save("output/figures/preprint_published_frequency_odds.svg")
```
ajlee21, line 213: Interesting, the most enriched for either preprint or publication seems to be these nonsensical letters. But then if you lower your enrichment score looks like there terms for experimental design in the published works and a lot about the covid19 outbreak in the preprints. I wonder what your takeaway from this plot is

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
```diff
@@ -0,0 +1,603 @@
...
+polka_preprints_df = pd.read_csv("output/polka_et_al_biorxiv_embeddings.tsv", sep="\t")
+polka_preprints_df.head()
+
+pca_components = pd.read_csv(
+    Path("../pca_association_experiment/output/word_pca_similarity/pca_components.tsv"),
+    sep="\t",
+)
+pca_components.head()
+
+# ## PCA Components
```
ajlee21, line 236: So here you're compressing your 300-dim word embedding into 50 PCs?

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
```diff
@@ -0,0 +1,603 @@
...
+    .assign(label="biorxiv")
+)
+# biorxiv_pca_sim_df.to_csv("output/polka_pca_enrichment.tsv", sep="\t")
+biorxiv_pca_sim_df.head()
+
+# ## PC Regression
+
+# ### Logistic Regression
+
+# Goal here is to determine if we can figure out which PCs bisect the bioRxiv subset against Polka et al.'s subset. Given that their dataset is only 60 papers we downsampled our dataset to contain only 60 papers.
```
ajlee21: What do you mean by "bisect" here? My mind is going to geometry:) and I don't think thats your intention

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
```diff
@@ -0,0 +1,603 @@
...
+model_weights_df = pd.DataFrame.from_dict(
+    {
+        "weight": model.coef_[0],
+        "pc": list(range(1, 51)),
+    }
+)
+model_weights_df["pc"] = pd.Categorical(model_weights_df["pc"])
+model_weights_df.head()
+
+g = (
```
ajlee21, line 317: This is how much each PC contributes to prediction of biorxiv vs pmcoa?

ajlee21: I wonder what PC 23, 50 are

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
```diff
@@ -0,0 +1,603 @@
...
+
+# + geom_point(data=subset_df, aes(x=dim1, y=dim2, colour = "red"))
+# )
+# print(g)
+# -
+# ## Publication Time Analysis
+
+# ### Get publication dates
+
+url = "https://api.biorxiv.org/pub/2019-11-01/3000-01-01/"
```
ajlee21, line 387: Is this highlighting where the polka et. al. papers on in your preprint landscape?

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
```diff
@@ -0,0 +1,603 @@
...
+    [
+        biorxiv_published_distances["version_count"].min(),
+        biorxiv_published_distances["version_count"].max(),
+    ]
+)
+y_line = x_line * results_2.slope + results_2.intercept
+
+# Graph here?
+plt.figure(figsize=(8, 5))
+g = sns.violinplot(
```
ajlee21, line 497: This makes sense. The longer it takes to publish the more preprint versions you would have. Is this serving as a positive control? I wonder if it would make sense to look at time to publish vs some other parameters?

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
```diff
@@ -0,0 +1,603 @@
...
+
+x_line = np.array(
+    [
+        biorxiv_published_distances["doc_distances"].min(),
+        biorxiv_published_distances["doc_distances"].max(),
+    ]
+)
+y_line = x_line * results_2.slope + results_2.intercept
+
+# graph here?
```
ajlee21, line 563: I think this is saying the same thing as  above, that the longer it takes to publish, the more changes your preprint will undergo

ajlee21 (approved):
All the code looks good to me. Just some conceptual questions as usual. 

One minor refactoring comment, that doesn't need to be addressed here. It seems like you're doing a lot of different analyses in notebook 3. I wonder if it would help to split those out?

biorxiv/polka_subset_experiment/01_download_polka_subset_experiment.py
danich1: Ah good call!

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
danich1: In general yes, but token is an umbrella term as it encompasses all forms of a word (tenses plus root). Lemmas are words that are in base form only. (E.g. walking -> walk) In respect to tokens "walking" and "walk" are unique entities.

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
danich1: Ah I see your confusion. That's bad nomenclature on my end. I'll fix that so it makes more sense.

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
danich1: Think about a contingency table for randomized control trials. The groups are preprints and published papers and the values are token counts. To reinforce the point column descriptions are below:

| Column | Description|
| --- | --- |
| Lemma | is the token being analyzed |
| corpus_one_a | the frequency of the token in question for the first corpus (in this case Polka et al subset) |
| corpus_one_b | the total number of tokens within Polka et al subset |
| corpus_two_c | the frequency of the token in question for the second corpus (in this case the published versions of Polka et al's subset) |
| corpus_two_d | the total number of tokens within published version |
| Log-likelihood | This is a function that states how well the above counts fit a contingency table model. higher the number better the fit. |
| Odds Ratio | This calculates which group a token is more associated with. It is a metric from 0 to infinity where numbers less than one indicate association with published papers, numbers greater than one indicate association with polka subset and numbers exactly at one indicate equal association across groups. |

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
danich1: These are 95% confidence intervals for the calculated odds ratio.

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
danich1: Heh. No can't say nonsensical, more like paper sections or gene mentions. The other letters are probably elements of equations as preprints within this subset are main focused around COVID-19 (Jan 2020-Dec 2020). One takeaway from this graph is that published version contains more medical trials and experiments in relation to the virus while preprints (in bioRxiv) are focused on functional experiments/epidemiological findings on the COVID-19 virus.

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
danich1: No. I'm using my 50 principal components I calculated within bioRxiv to see which PCs are associated with preprints within Polka's subset. (I'm trying to contextualize their work within mine.)

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
danich1: Bisect was suppose to mean separate Polka subset from random sample of bioRxiv preprints. I'll find a better word.

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
danich1: >This is how much each PC contributes to prediction of biorxiv vs pmcoa?

Correct. The idea was to see if logistic regression would be able to predict which preprint belongs to biorxiv and polka's subset using PC's similarity as features. By doing this we wanted to see which PCs are associated with this subset.

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
danich1: >Is this highlighting where the polka et. al. papers on in your preprint landscape? 

Yes. The idea is to see how their preprints are spread out.

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
danich1: >This makes sense. The longer it takes to publish the more preprint versions you would have. Is this serving as a positive control? I wonder if it would make sense to look at time to publish vs some other parameters?

Not positive control. More like we wanted to show that preprints within their subset are being published at a faster rate than average. (Hence the red does below the regression line.)

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
danich1: >I think this is saying the same thing as above, that the longer it takes to publish, the more changes your preprint will undergo

Yup. That's the main point. We are adding this subset on top to give the "future reader" some context as to where preprints within the subset are in respect to the bigger picture. By the time a reader sees this figure they would have already known the message you just stated above.

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
```diff
@@ -0,0 +1,602 @@
...
+        ]
+    )
+    .query(f"token not in {stop_word_list}")
+    .groupby("token")
+    .agg({"count": "sum"})
+    .reset_index()
+    .sort_values("count", ascending=False)
+)
+
+published_count_df.head()
```
ben-heil, line 145: Is there a reason that two of the lemmas are blank here? Are they whitespace

image

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
```diff
@@ -0,0 +1,602 @@
...
+    + p9.annotate("text", label="Preprint Enriched", x=-5, y=38.5, size=12, alpha=0.7)
+    + p9.theme_seaborn(context="paper", style="ticks", font_scale=1.1, font="Arial")
+    + p9.theme(
+        figure_size=(10, 6),
+        panel_grid_minor=p9.element_blank(),
+    )
+    + p9.labs(y=None, x="Preprint vs Published log2(Odds Ratio)")
+)
+g.save("output/figures/preprint_published_frequency_odds.svg")
+g.save("output/figures/preprint_published_frequency_odds.png", dpi=250)
```
ben-heil, line 214: Are the arrows in this figure backwards? Looking at the previous table `s.` is found 25 times in `corpus_one_a` and 145 times in `corpus_two_b`. If corpus one is preprints and corpus two is published documents, then `s.` would be enriched in publications, not preprints

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
```diff
@@ -0,0 +1,602 @@
...
+cbar = plt.colorbar()
+_ = cbar.ax.set_ylabel("count", rotation=270)
+_ = ax.scatter(
+    polka_published_preprint_df["doc_distances"],
+    polka_published_preprint_df["days_to_published"],
+    c="red",
+    s=6,
+)
+plt.savefig("output/figures/article_distance_vs_publication_time_hex.svg", dpi=250)
+plt.savefig("output/figures/article_distance_vs_publication_time_hex.png", dpi=250)
```
ben-heil, line 592: I wonder if the outliers on this plot are the same as the outliers in the SAUCIE plot. Maybe some types of preprint just have a lot of versions (though I guess with n ~ 2 you couldn't really say anything for sure)

ben-heil (commented):
Looks good! Be sure to check the arrow direction comment. I'm probably wrong about them being backwards, so once you've looked at that consider this review to be an "Approve"

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
danich1: >Is there a reason that two of the lemmas are blank here? Are they whitespace

Yup those are spaces/tabs.

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
danich1: Great Catch!

biorxiv/polka_subset_experiment/03_rerun_analyses_with_added_subset.py
danich1: Only one is a true outlier on both plots. (doi=10.1101/2020.02.27.967760) If more papers were manually annotated in the subset that would have been an interesting trend.

PR #30, opened 2021-04-21
Rerun of document vector generation 1/n (n=2🤞🏾)

danich1:
This PR is a result of the conversation contained in [greenelab/preprint-similarity-search#81](https://github.com/greenelab/preprint-similarity-search/pull/81). It was suggested that changing how document vectors were generated would improve results. Based on this suggestion, I needed to regenerate results given the new parser. 

I attempted to make this PR into multiple smaller versions; however, github likes to accumulate changes when one doesn't switch to main branch before creating a new branch. Hence this large PR now. Good news: Since results only changed slightly given this whole rerun debacle. You only need to review two files: annorxiver_modules/document_helper.py and  annorxiver_modules/pca_plot_helper.py. Everything else is just data changes.


annorxiver_modules/document_helper.py
```diff
@@ -1,10 +1,15 @@
+import csv
+import lzma
 import re
 
-from gensim.parsing.preprocessing import remove_stopwords
```
ajlee21, line 3: Very high level question. What was out of date in your current parser?

annorxiver_modules/document_helper.py
```diff
@@ -115,6 +136,102 @@ def generate_doc_vector(model, document_path, xpath, filter_tags=filter_tag_list
     return []
 
 
+def generate_doc_vector_parallel(
+    model, xpath, document_queue, vector_queue, filter_tags=filter_tag_list
+):
+    """
+    This method is designed to construct document vectors for a given xml document.
+    Every document has specific tags that are striped in order to have accurate embeddings
```
ajlee21, line 146: What are these tags? Would it help to have an example in this description?

annorxiver_modules/document_helper.py
```diff
@@ -115,6 +136,102 @@ def generate_doc_vector(model, document_path, xpath, filter_tags=filter_tag_list
...
+                    lambda tok: tok.lemma_ in model.wv
+                    and tok.lemma_ not in nlp.Defaults.stop_words,
+                    nlp(all_text),
+                ),
+            )
+        )
+
+        word_vectors += [model.wv[text] for text in all_tokens]
+
+        # skips weird documents that don't contain text
```
ajlee21, line 186: Are these weird documents those very old ones that are images and not text? Have you run into other strange cases - i'm thinking about those older papers that published the protein sequences within the text of the paper for instance. I guess these would just be outliers. I'm not sure if it would throw off your embedding

annorxiver_modules/document_helper.py
```diff
@@ -115,6 +136,102 @@ def generate_doc_vector(model, document_path, xpath, filter_tags=filter_tag_list
...
+                    document_path[1],
+                    np.stack(word_vectors).mean(axis=0),
+                )
+            )
+
+    # Poison Pill to end writer
+    vector_queue.put(None)
+
+
+def write_document_vector_parallel(vector_queue, file_path_name, dim, n_jobs):
```
ajlee21, line 200: Maybe add a description for this function

ajlee21 (approved):
This looks good to me. I just had a few conceptual questions

annorxiver_modules/document_helper.py
danich1: Nothing wrong with using gensim's parser for removing stopwords; however, I switched to using spaCy for processing and they provide their own stopwords. To be consistent I switched.

annorxiver_modules/document_helper.py
danich1: For bioRxiv's xml documents these tags include extraneous text that are not conducive for document parsing (e.g. figures, metadata etc.). I'll put an example.

annorxiver_modules/document_helper.py
danich1: >Are these weird documents those very old ones that are images and not text?

You are correct. Those weird documents are just image scans of old papers focused on DNA/protein sequences. They don't affect results as they would appear as empty vectors that I'm skipping.

annorxiver_modules/document_helper.py
danich1: Shoot I thought I wrote a description for this function. Good catch!

annorxiver_modules/document_helper.py
```diff
@@ -115,6 +136,102 @@ def generate_doc_vector(model, document_path, xpath, filter_tags=filter_tag_list
...
+            vector_queue.put(
+                (
+                    document_path[0],
+                    document_path[1],
+                    np.stack(word_vectors).mean(axis=0),
+                )
+            )
+
+    # Poison Pill to end writer
+    vector_queue.put(None)
```
ben-heil, line 197: Is the None value in document_queue already in the queue when it is passed into the function? Should you add that assumption to the function comments to avoid infinite looping? Or does the queue just return None when it runs out of items?

ben-heil (approved):
Looks good!

annorxiver_modules/document_helper.py
danich1: >Is the None value in document_queue already in the queue when it is passed into the function? 

Actually no the None value gets passed once all the documents have been added to the document queue. Regarding parallel processing, timing is key. If I passed None in first then the loop will never run as it will prematurely close/

>Should you add that assumption to the function comments to avoid infinite looping? Or does the queue just return None when it runs out of items?

I can add the assumption to clarify the confusion. Queue actually returns an exception if it is empty so a better way is to add None to the queue.

PR #31, opened 2021-05-04
Rerun part2

danich1:
This PR follows up on #30. No review needed as its just the figures being updated.


ajlee21 (approved):
Looks good to me. From what I can remember the plots look very similar to your previous run which is good.

Just a couple of conceptual questions:
* It looks like your paper-paper score vs centroid score are performing very similarly, so you're using the centroid to predict journals because its faster and has similar performance to paper-paper?
* What does your saucie hyperparams plot show? I'm not very familiar with saucie so I wasn't sure what the different lambda values meant.

danich1:
>It looks like your paper-paper score vs centroid score are performing very similarly, so you're using the centroid to predict journals because its faster and has similar performance to paper-paper?

Using centroid analysis to prevent over-representation of journals that like to publish a lot of papers. Its an alternative way to show similar journals along with avoiding publication bias.

>What does your saucie hyperparams plot show? I'm not very familiar with saucie so I wasn't sure what the different lambda values meant.

So Saucie is a model designed for clustering RNA-seq data. lambda b - is the batch correction parameter, lambda c is penalty for the number of clusters and lambda d is a penalty term for how far apart each cluster are from each other. Basically only use two parameters as I don't need to batch correct papers :joy:. Basic idea is to see if SAUCIE can cluster papers based on their journal representation. Looking at the graphs the model definitely can!

ajlee21:
> > It looks like your paper-paper score vs centroid score are performing very similarly, so you're using the centroid to predict journals because its faster and has similar performance to paper-paper?
> 
> Using centroid analysis to prevent over-representation of journals that like to publish a lot of papers. Its an alternative way to show similar journals along with avoiding publication bias.
> 
This logic definitely makes sense for using the centroid approach. I guess a followup question is: is this figure meant to argue for using centroid vs paper-paper?

danich1:
>This logic definitely makes sense for using the centroid approach. I guess a followup question is: is this figure meant to argue for using centroid vs paper-paper?

No. The figure is just to show how the centroid classifier performs along with the paper-by-paper. Just to say if we look at journals alone performance doesn't decrease too much.

PR #34, opened 2021-08-31
Article distance abstract only

danich1:
A reviewer requested that I check to see if my preprint-published matcher works by using abstracts only rather than full text.  Another reviewer requested that I check the publication rate but ignore everything from 2019-2020 as those are the most recent years at time of submission. 

So with those two facts the following notebooks should be reviewed:
1. 03_biorxiv_corrected_published_rate.ipynb - check the two figures on the bottom
2. 04_biorxiv_article_distances_abstract_only.ipynb  - figures within this notebook. The scatter plot shows full text distances against abstract only distances to see how well both approaches agree.


biorxiv/article_distances/03_biorxiv_corrected_published_rate.py
```diff
@@ -185,9 +185,9 @@
     .groupby("pub_month")
```
ajlee21, line 186: Is there a reason you're replacing the month here? Maybe add a comment would be helpful here

biorxiv/article_distances/03_biorxiv_corrected_published_rate.py
```diff
@@ -292,3 +292,82 @@
 g.save("output/figures/publication_rate_rerun.svg")
 g.save("output/figures/publication_rate_rerun.png", dpi=250)
 print(g)
+
+# # Plot Publication Rate
```
ajlee21, line 297: Looks like rate = published/posted = number of papers published that month / ??? I guess I'm not sure what the posted means?

ajlee21: Also is your label where the data is coming from? So you have different data sources for different time points?

biorxiv/article_distances/03_biorxiv_corrected_published_rate.py
```diff
@@ -292,3 +292,82 @@
...
+print(f"Overall proportion published: {published/posted:.4f}")
+# -
+
+color_mapper = {
+    "2018": "#a6cee3",
+    "2020ML": "#33a02c",
+    "2020": "#1f78b4",
+}
+
+g = (
```
ajlee21, line 326: I would add a title so that you know which one truncated. The trends look the same as expected, but the overall publications is higher after your truncate since 2019 will include many papers that are not published yet, is that right?

ajlee21: Is there a reason you still see 2019 in the truncated figure?

ajlee21 (commented):
Here are some initial comments. Will finish reviewing after you reupload the 04 notebook, which is missing the bottom figure.

ajlee21 (approved):
So this is the distance between published and preprint using abstract embeddings vs full text embeddings? And the distances between published and preprints are very close using either embedding except for a few outliers -- any thoughts on what those are?

biorxiv/article_distances/04_biorxiv_article_distances_abstract_only.py
```diff
@@ -0,0 +1,239 @@
...
+            }
+        )
+    )
+    + p9.aes(x="label", y="distance")
+    + p9.geom_violin(fill="#a6cee3")
+    + p9.labs(x="Document Pair Groups", y="Euclidean Distance")
+    + p9.theme_seaborn(context="paper", style="ticks", font="Arial", font_scale=1.35)
+)
+g.save("output/figures/biorxiv_article_distance_abstract_only.svg", dpi=250)
+g.save("output/figures/biorxiv_article_distance_abstract_only.png", dpi=250)
```
ben-heil, line 201: The distances for the abstract_only plots seem to be larger than the ones when looking at the entire article. Is that an artifact of normalization in different text sizes? Or do abstracts change more than the full text when publishing?

I guess since the distance to random articles is also different, would it be possible to normalize based on the mean random article distance so the two plots are more comparable?

biorxiv/article_distances/04_biorxiv_article_distances_abstract_only.py
```diff
@@ -0,0 +1,239 @@
...
+    + p9.labs(x="Full Text Distance (-log 10)", y="Abstract Only Distance (-log 10)")
+    + p9.theme_seaborn(context="paper", style="ticks", font="Arial", font_scale=1.35)
+)
+g.save("output/figures/biorxiv_full_text_v_abstract_only.svg", dpi=250)
+g.save("output/figures/biorxiv_full_text_v_abstract_only.png", dpi=250)
+print(g)
+
+plot_df.sort_values("abstract_only_distance_log10", ascending=False).head(10)
+
+# The negative values for abstract only consist of high confidence scores for a true match. The table for the first few entries have an abstract only distance of practically zero. This means the log version will be highly negative.
```
ben-heil: Do you have an intution for what the cluster at the bottom is? Are those just articles whose abstract changed completely when they were published?

ben-heil (approved):
Looks good!

biorxiv/article_distances/03_biorxiv_corrected_published_rate.py
danich1: So a small handful of preprints were "posted" during October of 2013; however bioRxiv didn't really start until November 2013. So I'm updating their posting to be concurrent with bioRxiv's true start month. Adding a comment to reflect this point.

biorxiv/article_distances/03_biorxiv_corrected_published_rate.py
danich1: >Looks like rate = published/posted = number of papers published that month / ??? I guess I'm not sure what the posted means?

So posted means the date at which bioRxiv accepts a preprint and hosts it on their site. We are looking for the amount of preprints that are published compared to the amount of preprints posted for each month.

>Also is your label where the data is coming from? So you have different data sources for different time points?

Yes. The light blue is recapitulating a study that was previously done looking at the publication rate of preprints. The dark blue is our updated rate to show how publication changed over time. (Can't freeze papers from being published, so we provided an updated plot.) The dark green plot is our suggested correction when one accounts for missing links within bioRxiv as the matching process isn't the best.

biorxiv/article_distances/03_biorxiv_corrected_published_rate.py
danich1: >I would add a title so that you know which one truncated. The trends look the same as expected, but the overall publications is higher after your truncate since 2019 will include many papers that are not published yet, is that right?

Yup. You are correct. One of my reviewers wanted to see how the rate looked if you didn't take 2019+ into account. Hence my cutoff plot.

>Is there a reason you still see 2019 in the truncated figure?

To show the future of preprints being posted onto bioRxiv. I didn't want to remove that part of the figure as its removal would open a door for a follow up request asking about 2019-2020 posted rate.

biorxiv/article_distances/04_biorxiv_article_distances_abstract_only.py
danich1: This description is wrong and will be changed. Turns out I took the negative log of the distances so what I wrote is misleading. 

Each point within this plot are preprints matched with their corresponding published article. The intuition here is to see if abstract distances correlate with full text distances. If it does, then that would mean abstracts alone can be used to match documents. The blob below is majority of the articles while the outliers above are papers whose abstracts didn't change very much or at all. (will manually confirm) Long story short: I'll be updating this figure and description so everything makes more sense.

biorxiv/article_distances/04_biorxiv_article_distances_abstract_only.py
danich1: >The distances for the abstract_only plots seem to be larger than the ones when looking at the entire article. Is that an artifact of normalization in different text sizes? Or do abstracts change more than the full text when publishing?

Based on [this paper](https://www.biorxiv.org/content/10.1101/2021.02.20.432090v2.full.pdf), I'm inclined to say that this is an artifact from the approach I'm using to generate document vectors. Full text has a lot more words than abstracts, so it would be more agnostic to small changes. Hence lower document distances.

>I guess since the distance to random articles is also different, would it be possible to normalize based on the mean random article distance so the two plots are more comparable?

Hmm.. Can't think of any methods off the top of my head, but will see more into this question

danich1:
>So this is the distance between published and preprint using abstract embeddings vs full text embeddings? And the distances between published and preprints are very close using either embedding except for a few outliers -- any thoughts on what those are?

So the large outliers using abstract only are papers that have undergone an abstract structural change or large phrases within the abstract has changed. Did a manual review on some of the abstract only outliers and found these to be the case. As for full text I hypothesize it's a simple matter of large changes with the published paper itself due to peer review. This leads to outliers within my dataset in respect to full text.

biorxiv/article_distances/04_biorxiv_article_distances_abstract_only.py
danich1: > Hmm.. Can't think of any methods off the top of my head, but will see more into this question

To follow up on my own comment. I couldn't think of any normalization tricks to make abstracts directly comparable with full text; however, I think the histogram plot gives the desired message that abstracts alone are a viable way to match documents (if you accept the high sensitivity to small changes as a caveat).

PR #35, opened 2021-09-07
Corpora Comparison Reviewer Request

danich1:
A reviewer asked what the results would be if I removed special characters from my token odds ratio analysis. 
This PR address that question and updates figures to have larger font for manuscript changes.

Only review needed is for this file that contains results: 
1. 05_figure_generator_reviewer_request.ipynb

The token figures with NYTAC are a bit malformed. Don't worry about that as I'm only reporting bioRxiv vs PMCOA, preprint vs published and the first KL divergence plot.


biorxiv/corpora_comparison/05_figure_generator_reviewer_request.py
```diff
@@ -0,0 +1,957 @@
...
+        alpha=1,
+        arrow=p9.arrow(length=0.1),
+    )
+    + p9.annotate("text", label="PMC Enriched", x=-1.5, y=38.5, size=18, alpha=0.7)
+    + p9.theme_seaborn(context="paper", style="ticks", font_scale=1.8, font="Arial")
+    + p9.theme(
+        figure_size=(11, 8.5),
+        panel_grid_minor=p9.element_blank(),
+    )
+    + p9.labs(y=None, x="bioRxiv vs PMC log2(Odds Ratio)")
```
ben-heil: The bars for et, al, and genes seem to be missing from this plot. Maybe because their error bar is too thin?

Also, is there a difference in how you do stemming or tokenization between the analyses with and without special characters? It seems weird that one has both "patient" and "patients" in its top pubmed terms while the other only has "patients"

ben-heil (approved):

biorxiv/corpora_comparison/05_figure_generator_reviewer_request.py
danich1: >The bars for et, al, and genes seem to be missing from this plot. Maybe because their error bar is too thin?

Good question. Turns out plotnine likes to remove bars if the y-axis ticks become too large. Have updated that fix that issue.

>Also, is there a difference in how you do stemming or tokenization between the analyses with and without special characters? It seems weird that one has both "patient" and "patients" in its top pubmed terms while the other only has "patients"

Normally there is no difference when removing special characters. When I performed this analysis I didn't  have spacy lemmatize each token. Therefore, "patient" and "patients" are considered separate tokens and have equal chance of showing up for this supplemental part.

PR #36, opened 2021-09-07
Testing of Doc2Vec and Word2Vec Pretrained models

danich1:
A reviewer asked my why I didn't use doc2vec nor pretrain word2vec before conducting my experiments using document vectors.
Main answer is that doc2vec is hard to interpret and that pretraining doesn't add much to performance in terms of visualization the landscape. 

Main file to review is:
1. 07_biorxiv_visualize_embeddings_pretrained.ipynb

This file contains umap and PCA comparisons of the different methods. If you'd like to look at other notebooks go ahead, but they mainly focus on training doc2vec or pretraining word2vec.


biorxiv/word_vector_experiment/07_biorxiv_visualize_embeddings_pretrained.py
```diff
@@ -0,0 +1,195 @@
...
+import re
+
+import pandas as pd
+import plotnine as p9
+# -
+
+journal_map_df = pd.read_csv(
+    "../exploratory_data_analysis/output/biorxiv_article_metadata.tsv", sep="\t"
+)
+journal_map_df.head()
```
ajlee21, line 33: What do the values in the "heading" column indicate?

biorxiv/word_vector_experiment/07_biorxiv_visualize_embeddings_pretrained.py
```diff
@@ -0,0 +1,195 @@
...
+                "pca1": "last",
+                "pca2": "last",
+                "category": "last",
+            }
+        )
+        .reset_index(drop=True)
+    ),
+}
+
+# # UMAP of the Documents
```
ajlee21, line 133: Would you mind adding a brief description about what the difference is between these visualizations:
So you're visualizing articles from different categories using 1) document embeddings, 2) doc2vec embeddings or 3) word2vec embeddings. What is the difference between in how these methods embed the document? Looks like doc2vec has less separation but word2vec is similar to your original embedding. I'm not sure how much the dimension matters either.

biorxiv/word_vector_experiment/07_biorxiv_visualize_embeddings_pretrained.py
```diff
@@ -0,0 +1,195 @@
...
+g = (
+    p9.ggplot(biorxiv_umap_models_latest["doc2vec"])
+    + p9.aes(x="umap1", y="umap2", color="factor(category)")
+    + p9.geom_point()
+    + p9.labs(title="UMAP of BioRxiv (Doc2vec Word dim: 300)", color="Article Category")
+)
+g.save("output/embedding_output/umap/figures/biorxiv_umap_300_doc2vec.png", dpi=500)
+print(g)
+
+g = (
```
ajlee21, line 159: I'm having a little trouble interpreting these plots, particularly what the difference between UMAP vs PCA in this case. It looks like you're able to capture more of the difference between article categories in UMAP vs PCA, though I'm not sure why this is the case. Also in general it feels like articles are pretty squished towards 0 in both visualizations - is this expected?

ajlee21 (approved):
Nice experiment!

Just a few conceptual questions. As a non-NLP person, is the metric to compare models/embeddings to use this type of visualization? Or is there some other metric to compare how "well" a model captures the information in the text?

mthielk:
man, they're really making you do a lot more work...
> A reviewer asked my why I didn't use doc2vec nor pretrain word2vec before conducting my experiments using document vectors.
Main answer is that doc2vec is hard to interpret and that pretraining doesn't add much to performance in terms of visualization the landscape.

I would only imagine pretraining word2vec would be useful if you didn't have enough text data for the embeddings to stabilize. Is there even an agreed upon way to know when you're done training your word vectors? I've always just seen train for n epochs and see if you get any signal. What other corpus are you using for pre-training? 

If you wanted a quantitate metric you might try looking at how separable the categories are in each embedding. i.e. the performance of a a multiclass logistic regression (or even xgboost if you want a nonlinear comparison) trained to predict the category from the embedding. Probably not necessary but might protect you from some of the shadows Lior has been tossing on twitter.

biorxiv/word_vector_experiment/07_biorxiv_visualize_embeddings_pretrained.py
danich1: This "heading" column represents the type of result a preprint is reporting. Majority of preprints report a new result, but some are contradictory or confirmatory research.

biorxiv/word_vector_experiment/07_biorxiv_visualize_embeddings_pretrained.py
danich1: >So you're visualizing articles from different categories using 1) document embeddings, 2) doc2vec embeddings or 3) word2vec embeddings. What is the difference between in how these methods embed the document? Looks like doc2vec has less separation but word2vec is similar to your original embedding. I'm not sure how much the dimension matters either.

I'll update the notebook with a description, but to answer your question: 
1) This is my original approach where I'm taking the average of every word vector present within a document. 
2) Doc2vec generates a document vector and word vectors at the same time. It's an updated word2vec model where every sentence is tagged with a document id and each document has their own randomly initialized vector. As the neural network processes each sentence it updates the corresponding document vector. So once training is done, each document has a "representation" vector and every word has their own "representation" vector.  
3) This approach is the same as (1) with the exception that the model is initialized with an already trained word2vec network. This is called pretraining. Once each vector has been initialized proceed to train word2vec normally. Within the NLP field pretraining can help when your corpus is not sufficiently large enough to train.

biorxiv/word_vector_experiment/07_biorxiv_visualize_embeddings_pretrained.py
danich1: >I'm having a little trouble interpreting these plots, particularly what the difference between UMAP vs PCA in this case. It looks like you're able to capture more of the difference between article categories in UMAP vs PCA, though I'm not sure why this is the case. Also in general it feels like articles are pretty squished towards 0 in both visualizations - is this expected?

The difference between UMAP and PCA are the way the low dimensional space is constructed. UMAP is non-linear and PCA is linear. The idea here is to visually show the differences between doc2vec, word2vec and pretraining. Doc2vec clusters documents a whole lot differently than word2vec, while pretraining doesn't change anything (argument against using a pretrained model for my response to reviewers). 

In terms of the 0 squishing I'd speculate that a lot of preprints contain similar concepts (e.g. sequencing or other general approaches related to that idea)  that are captured by these latent embeddings from word2vec/doc2vec. Is there any meaning in this? Probably not as this is just for visual purposes rather than in depth meaning.

danich1:
>Just a few conceptual questions. As a non-NLP person, is the metric to compare models/embeddings to use this type of visualization? Or is there some other metric to compare how "well" a model captures the information in the text?

In this case there isn't a quantitative way to compare these visualizations to my knowledge. Normal routine is to see difference in performance for downstream analyses (e.g. Document classification, sentiment analysis, etc.) In that space the more advanced neural networks win as they get boost in performance. In regards to interpretability that still has a long way to go, but there is research out there trying to break open the black box.

danich1:
>I would only imagine pretraining word2vec would be useful if you didn't have enough text data for the embeddings to stabilize. Is there even an agreed upon way to know when you're done training your word vectors? I've always just seen train for n epochs and see if you get any signal. What other corpus are you using for pre-training?

There isn't an agreed upon way to know you are done training without seeing how performance changes for certain tasks. This [paper](https://aclanthology.org/W16-2922.pdf) attempts to answer what are effective parameters when training word2vec; however, the number of epochs wasn't explored.  So it's mainly up to the user to determine best results. Was using Google's news 300 dim dataset as it seemed the closest towards the model I was using. As expected it didn't significantly change the outcome of the embedding space.

>If you wanted a quantitate metric you might try looking at how separable the categories are in each embedding. i.e. the performance of a a multiclass logistic regression (or even xgboost if you want a nonlinear comparison) trained to predict the category from the embedding. Probably not necessary but might protect you from some of the shadows Lior has been tossing on twitter.

Good idea. Will keep in mind if I get push back.

biorxiv/word_vector_experiment/07_biorxiv_visualize_embeddings_pretrained.py
ajlee21: Ah I see. So 2 and 3 are updating an embedding based on that task of finding likelihood that the word "love" fills in the sentence:  "I ____ cats"? Thanks for the clarification

biorxiv/word_vector_experiment/07_biorxiv_visualize_embeddings_pretrained.py
ajlee21: I see. I think I was trying to compare the results using UMAP vs PCA. It does look like you have the same takeaway message from either method, I was just noticing that you got a bit better separation between clusters using UMAP which is where I was coming from.

biorxiv/word_vector_experiment/07_biorxiv_visualize_embeddings_pretrained.py
danich1: >Ah I see. So 2 and 3 are updating an embedding based on that task of finding likelihood that the word "love" fills in the sentence: "I ____ cats"? Thanks for the clarification

All three do actually.  2 has an extension where a document vector gets updated along with word vectors, while 3 is exactly the same model as 1 just vectors are instantiated with another model rather than randomly initialized.

biorxiv/word_vector_experiment/07_biorxiv_visualize_embeddings_pretrained.py
danich1: >I see. I think I was trying to compare the results using UMAP vs PCA. It does look like you have the same takeaway message from either method, I was just noticing that you got a bit better separation between clusters using UMAP which is where I was coming from.

Yeah. The beauty of UMAP. It is better at separating clusters than PCA due to its non-linearity. I'll update my notebook description to be clear on the intent for this analysis.

PR #37, opened 2021-09-09
Publication time reviewer request

danich1:
This PR is designed to upload results to answer a reviewer's question on if text differences increase with the number of different preprint versions. Also the notebook outputs an excel file containing preprints published pairs along with their corresponding publication information.

Main files to be reviewed:
1. 03_publication_delay_experiment_reviewer_request.ipynb
2. version_count_doc_distances.png


biorxiv/publication_delay_experiment/03_publication_delay_experiment_reviewer_request.py
```diff
@@ -0,0 +1,122 @@
...
+#       jupytext_version: 1.9.1+dev
+#   kernelspec:
+#     display_name: Python [conda env:annorxiver]
+#     language: python
+#     name: conda-env-annorxiver-py
+# ---
+
+# # Measure the Difference between Preprint-Published similarity and Published Articles
+
+# This notebook is designed to output an excel file that contains all preprint published pairs and their respective publication information.
```
ajlee21: Maybe also add the analysis about preprint version vs doc distance here too

biorxiv/publication_delay_experiment/03_publication_delay_experiment_reviewer_request.py
```diff
@@ -0,0 +1,122 @@
...
+    )
+    + p9.theme_seaborn(style="white", context="notebook")
+)
+g.save("output/version_count_doc_distances.svg")
+g.save("output/version_count_doc_distances.png", dpi=600)
+print(g)
+
+# # Output published dates to Excel
+
+# Reviewer asked if manually pursuing preprints that take longer to publish would produce any interesting results. Great question, but not enough time to go into that; however, providing a supplementary file for others to look into could provide an in depth answer.
```
ajlee21, line 93: nice!

biorxiv/publication_delay_experiment/03_publication_delay_experiment_reviewer_request.py
```diff
@@ -0,0 +1,122 @@
...
+        x=9,
+        y=7.5,
+        size=13,
+        color="#1f78b4",
+    )
+    + p9.labs(
+        x="# of Preprint Versions",
+        y="Euclidean Distance of Preprint-Published Versions",
+    )
+    + p9.theme_seaborn(style="white", context="notebook")
```
ajlee21, line 83: What is the takeaway for this plot? It looks like if you remove the "outliers" that you might have a nice positive correlation between the distance and the number of versions, which I would intuitively expect more text changes with more version and thus a large difference between the final and original version. However, I'm not sure what these "outlier"-looking documents are so I wouldn't want to suggest removing them. I'm more curious what your thoughts are here.

ajlee21 (approved):
Really cool analysis. Just a couple of comments

biorxiv/publication_delay_experiment/03_publication_delay_experiment_reviewer_request.py
danich1: >What is the takeaway for this plot? It looks like if you remove the "outliers" that you might have a nice positive correlation between the distance and the number of versions, which I would intuitively expect more text changes with more version and thus a large difference between the final and original version. However, I'm not sure what these "outlier"-looking documents are so I wouldn't want to suggest removing them. I'm more curious what your thoughts are here.

So basically this plot shows a weak but positive relationship between number of preprint versions and document embedding distances. Based on the slope it says that for every preprint version the document distance changes by 0.028. So preprints are being changed with each new version. If you compared that slope to the distance between a random preprint and a published article (~ 6 distance units), then you can say that these "changes" are minor relative to a full article change. This goes against a previous assumption that each new preprint version should bring substantial changes, which is interesting. Overall, preprints do change with each new version. Just not to the extent as one, without prior knowledge about publishing and preprints, would expect.

PR #40, opened 2021-09-21
Preprint-Published Pairs Vector Distance Experiment

danich1:
This PR is based on comparing various distance metrics when determining preprint-published document pairs. Common approach in the field is to use cosine distance to determine "document similarity" as this metric is magnitude agnostic. For my paper I didn't use cosine distance I used pure euclidean distance. Based on this fact, a reviewer wanted to see if cosine distance makes a difference and results I found is that it does not for this particular task. 

I also viewed how abstracts only fairs against using full text. 
FYI: cosine proxy is basically using euclidean distance to mimic cosine distance [(L2 norm of vectors)](https://en.wikipedia.org/wiki/Cosine_similarity#%7F'%22%60UNIQ--postMath-00000011-QINU%60%22'%7F-normalised_Euclidean_distance)


biorxiv/article_distances/05_biorxiv_article_distances_cosine.py
```diff
@@ -0,0 +1,440 @@
...
+            }
+        )
+    )
+    + p9.aes(x="label", y="distance")
+    + p9.geom_violin(fill="#a6cee3")
+    + p9.labs(x="Document Pair Groups", y="Cosine Distance")
+    + p9.theme_seaborn(context="paper", style="ticks", font="Arial", font_scale=2)
+    + p9.theme(figure_size=(11, 8.5))
+)
+print(g)
```
ben-heil, line 386: It's interesting how much closer to zero the cosine distance is for preprint-published pairs. I guess it makes sense that people would use it more

biorxiv/article_distances/06_biorxiv_article_distances_cosine_abstract_tester.py
```diff
@@ -0,0 +1,423 @@
...
+
+plot_df = pd.DataFrame.from_records(data_rows)
+plot_df.head()
+
+g = (
+    p9.ggplot(plot_df, p9.aes(x="N", y="recall", color="model"))
+    + p9.geom_point()
+    + p9.labs(x="Top N predictions", y="Recall")
+)
+g.save("output/figures/abstract_vs_full_text_top_k_recall.png", dpi=600)
```
ben-heil, line 409: Having a auPRC of 1 for the top 10k predictions is pretty hard to argue with (I know that's not the point of these figures, but it is striking)

ben-heil (approved):
Looks good! Why do you use the proxy of the cosine similarity instead of the true cosine similarity? Is it that the dataset is too big to run cosine in a reasonable amount of time?

biorxiv/article_distances/06_biorxiv_article_distances_cosine_abstract_tester.py
```diff
@@ -0,0 +1,423 @@
...
+#       format_name: light
+#       format_version: '1.5'
+#       jupytext_version: 1.9.1+dev
+#   kernelspec:
+#     display_name: Python [conda env:annorxiver]
+#     language: python
+#     name: conda-env-annorxiver-py
+# ---
+
+# # Find published articles missing from bioRxiv
```
ajlee21: Maybe change the title here to reflect that you're looking at abstracts

biorxiv/article_distances/06_biorxiv_article_distances_cosine_abstract_tester.py
```diff
@@ -0,0 +1,423 @@
...
+    final_abstract_df
+)
+final_abstract_df.to_csv(
+    "output/annotated_links/article_distances_abstract_only.tsv", sep="\t", index=False
+)
+final_abstract_df.head()
+
+# # Distribution plot
+
+g = (
```
ajlee21, line 340: If this plot is the same as in the previous notebook, can you not just import it so you're not re-generating it again and thereby not duplicating code?

biorxiv/article_distances/06_biorxiv_article_distances_cosine_abstract_tester.py
```diff
@@ -0,0 +1,423 @@
...
+    )
+    + p9.aes(x="label", y="distance")
+    + p9.geom_violin(fill="#a6cee3")
+    + p9.labs(x="Document Pair Groups", y="Euclidean Distance")
+    + p9.theme_seaborn(context="paper", style="ticks", font="Arial", font_scale=2)
+    + p9.theme(figure_size=(11, 8.5))
+)
+print(g)
+
+g = (
```
ajlee21, line 357: This doesn't change your overall message, but I was just curious why there seem to be more outliers (i.e. the range of the y-axis is larger) comparing preprint abstract to published abstract but not as much using full text?

biorxiv/article_distances/05_biorxiv_article_distances_cosine.py
```diff
@@ -0,0 +1,440 @@
...
+
+g = (
+    p9.ggplot(plot_df, p9.aes(x="N", y="precision", color="distance"))
+    + p9.geom_point()
+    + p9.labs(x="Top N predictions", y="Precision")
+)
+g.save("output/figures/distance_metrics_top_k_precision.png", dpi=600)
+print(g)
+
+# Take Home Points:
```
ajlee21, line 436: SInce the distance metric doesn't seem to change your overall message, is the plan to use cosine distance since this is more standard? Also is there a reason you chose to use Euclidean in the first place, I forget

biorxiv/article_distances/06_biorxiv_article_distances_cosine_abstract_tester.py
```diff
@@ -0,0 +1,423 @@
...
+    p9.ggplot(plot_df, p9.aes(x="N", y="recall", color="model"))
+    + p9.geom_point()
+    + p9.labs(x="Top N predictions", y="Recall")
+)
+g.save("output/figures/abstract_vs_full_text_top_k_recall.png", dpi=600)
+print(g)
+
+g = (
```
ajlee21, line 412: Can you remind me of the task here. I remember that there are some pre-prints that are missing the link to their published counter part. So this task is trying to use the closest published document to fill in those missing links?

ajlee21 (approved):
Really nice concise experiment. Just a few additional question on top of what Ben has already noted

danich1:
>Looks good! Why do you use the proxy of the cosine similarity instead of the true cosine similarity? Is it that the dataset is too big to run cosine in a reasonable amount of time?

So cosine similarity (distance) is not a true distance metric as it violates the triangle inequality. Algorithms such as k-nearest-neighbors need to a true distance metric in order to operate efficiently (runtime not performance). Given this caveat, a proxy to the cosine distance can be used. I included it in this analysis to show that the proxy performs as well as the original metric.

biorxiv/article_distances/06_biorxiv_article_distances_cosine_abstract_tester.py
danich1: >If this plot is the same as in the previous notebook, can you not just import it so you're not re-generating it again and thereby not duplicating code?

Normally yes that is exactly what one should do. I left this in there as to have a sanity check for my distance code. The fact that the plot is exactly the same means I didn't mess anything up.

biorxiv/article_distances/06_biorxiv_article_distances_cosine_abstract_tester.py
danich1: >This doesn't change your overall message, but I was just curious why there seem to be more outliers (i.e. the range of the y-axis is larger) comparing preprint abstract to published abstract but not as much using full text?

Abstracts are short in nature. Given the small amount of sentences, their embedding distances are more sensitive to changes compared to full text (less words being incorporated).

biorxiv/article_distances/06_biorxiv_article_distances_cosine_abstract_tester.py
danich1: >Can you remind me of the task here. I remember that there are some pre-prints that are missing the link to their published counter part. So this task is trying to use the closest published document to fill in those missing links?

BioRxiv's matching approach isn't perfect and can result in preprints not being linked correctly. 
Our goal is to utilize this document embedding space to see if we can match preprints with their published counterparts. Which in turn lets us find missing links among the preprints without a link to their published version. Pairs that have a really low embedding distance can be considered the same paper.

biorxiv/article_distances/05_biorxiv_article_distances_cosine.py
danich1: >Since the distance metric doesn't seem to change your overall message, is the plan to use cosine distance since this is more standard? Also is there a reason you chose to use Euclidean in the first place, I forget

The opposite actually. The plan is to use these results to show that using a pure euclidean distance approach doesn't significantly affect performance. The main reason for Euclidean space was for runtime efficiency as Euclidean space runs faster than cosine distance (Euclidean distance is a true distance metric which is required by some algorithms).

ajlee21: That makes sense. Thank you for clarifying

PR #43, opened 2021-10-25
Plos final touches

danich1:
You dont have to in depth review. Just take a look at FIGURE_DATA_SOURCE.md. This file is needed for me to point readers towards my data files without the need to navigate my dense repository. Should be quick.


FIGURE_DATA_SOURCE.md
```diff
@@ -0,0 +1,91 @@
+# Figure Data Map
+
+## About
+This file is designed to point users and readers to datasets that were used to generate each figure piece.
+Each figure heading will have a table that specifies the following information: 
+- the panel piece label
```
ben-heil: I assume that a panel piece is a subarea in a figure, but I'm not sure. Can you add a sentence explaining what that is, or embed an example?

FIGURE_DATA_SOURCE.md
```diff
@@ -0,0 +1,91 @@
+# Figure Data Map
+
+## About
+This file is designed to point users and readers to datasets that were used to generate each figure piece.
```
ben-heil: You use "figure piece" here, "panel piece" in the next paragraph, and "panel" further down. If each is referring to the same thing, it'll be clearer if you stick to one name

FIGURE_DATA_SOURCE.md
```diff
@@ -0,0 +1,91 @@
...
+- [Figure Three](#figure-three)
+- [Figure Five](#figure-five)
+- [Figure S1](#figure-s1)
+- [Figure S2](#figure-s2)
+- [Figure S3](#figure-s3)
+- [Figure S4](#figure-s4)
+- [Figure S5](#figure-s5)
+- [Figure S6](#figure-s6)
+
+## Figure One
```
ben-heil, line 24: This table is really cool, I don't know that I've ever seen data + code + figures mapped explicitly like this before

ben-heil (approved):
Looks good, just one comment about consistency of a term

FIGURE_DATA_SOURCE.md
danich1: Glad you like it. It was my solution to prevent duplicating data within my repository or being forced to manually send data to the journal.

